Issue #26, opened 2014-12-08
Log noise from HTTPS certificate verification


addamh:
More of an annoyance than bug, but: 

starting a job with tower-cli with monitor results in a warning from python about certificate verification

is this on my end our yours? 

thanks in advance

```
Running script  : tower-cli job launch --job-template 5 --monitor
[workspace] $ /bin/sh -xe /tmp/hudson4436722897181643784.sh
+ tower-cli job launch --job-template 5 --monitor
/usr/lib/python2.6/site-packages/requests/packages/urllib3/connectionpool.py:734: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.org/en/latest/security.html
  InsecureRequestWarning)
/usr/lib/python2.6/site-packages/requests/packages/urllib3/connectionpool.py:734: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.org/en/latest/security.html
  InsecureRequestWarning)
/usr/lib/python2.6/site-packages/requests/packages/urllib3/connectionpool.py:734: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.org/en/latest/security.html
  InsecureRequestWarning)
/usr/lib/python2.6/site-packages/requests/packages/urllib3/connectionpool.py:734: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.org/en/latest/security.html
  InsecureRequestWarning)
/usr/lib/python2.6/site-packages/requests/packages/urllib3/connectionpool.py:734: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.org/en/latest/security.html
  InsecureRequestWarning)
/usr/lib/python2.6/site-packages/requests/packages/urllib3/connectionpool.py:734: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.org/en/latest/security.html
  InsecureRequestWarning)
Current status: pending
[...]

PR #29, opened 2014-12-16
Added 'verify_ssl' configuration flag.

wybczu:
Also suppress InsecureRequestWarning urllib3 warnings when verify_ssl == False. This PR solves  #26.


wybczu:
Hi @lukesneeringer,

This PR is related to cf37c28, but it suppress only `InsecureRequestWarning`. Can you take a look at this PR?

amokan:
this config feature would be helpful for me as well (cc @lukesneeringer @wybczu )

tristanbes:
Can you rebase it so an admin can merge this please ? it'd be great :-)

robbiet480:
:+1:. Annoying/scary to others that our SSL isn't passing, because we have an internal CA which our CI/CD doesn't recognize.

wybczu:
Also it is possible to specify custom [CA bundle](http://docs.python-requests.org/en/latest/api/#requests.request) via `verify_ssl` to verify SSL certificate.

wybczu:
On second thought this warning **should** be displayed. Not verifying ssl certificate is a big security issue, it makes MITM attacks easy :) I'm closing this pull request and I will create another one.

ccollicutt:
@wybczu What made you change your mind? What's wrong with turning off ssl warnings? If it's not default, then I'd like to be able to turn if off. What if the particular tower instance is just a test instance and we don't want to take the time to deploy proper certificates? Or what if it was a situation where I don't manage the service?

wybczu:
@ccollicutt as a user I would like to know when SSL verification is disabled (warnings should be displayed).

If you don't want to deploy proper cert you can always download tower's certificate (bash oneliner) and pass path to it e.g. via `ca_path` option. What do you think about such solution?

flmmartins:
Hi, was this merged?

I am trying to disable the verification it but no success.

PR #34, opened 2015-02-06
Remove egg_base=/tmp  in setup.cfg to allow installation on Windows machines via pip

schwartzmx:
This should not break other installations.  But currently installation via pip causes the error:

`error: error in 'egg_base' option: '/tmp' does not exist or is not a directory`

towards the end of the installation process on Windows machines.  This commit fixes the problem, and allows the installation to complete successfully on Windows.

Edit:  The only thing that will obviously not work (on Windows) is the `--global` config, since it looks for the file at `/etc/awx/`.

Questions, or concerns please let me know.

Thanks,

Phil

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,7 +1,3 @@
-[egg_info]
-egg_base = /tmp
-
-
 [nosetests]
 # don't wrap stdout/stderr
 nocapture=1


AlanCoding:
After reading up on the background of this, I see that the status-quo is less standard than the alternative. The only problem I ran into was that it dumps a directory ending in `.egg_info` into the tower-cli folder when running `pip install` or `setup.py install`. To deal with that, we just need another line in the `.gitignore` file. I see how it is better to avoid using a OS-specific temporary directory.

Thanks for the contribution. I'm going to branch off the current master and make those two changes while I close this PR.

PR #40, opened 2015-02-27
Add support for specifying tags

perzizzle:
Allow the specification of tags on a job. For example  tower-cli job launch --job-template=123 --tags=tag1,tag2,tag3. Tested with ansible tower 2.1.0


perzizzle:
I've tested this on ansible tower 2.1, but I don't have access to older versions.

PR #43, opened 2015-03-20
Ensure that --extra-vars file added via command line is added to the defa...

cedub:
...ult variables rather than overwriting them.

### DIFF
--- a/lib/tower_cli/resources/job.py
+++ b/lib/tower_cli/resources/job.py
@@ -74,7 +74,21 @@ class Resource(models.MonitorableResource):
         if extra_vars:
             if hasattr(extra_vars, 'read'):
                 extra_vars = extra_vars.read()
-            data['extra_vars'] = extra_vars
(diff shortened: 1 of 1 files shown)


cedub:
My understanding of "extra_vars" leads me to believe that this parameter should update the existing job template parameters rather than replacing them entirely.  I've updated the code as such.  If this is contrary to the intent of the project, I will just make changes locally.

Also, I've not been able to get the tests executing locally yet, so I apologize for pushing up a PR that hasn't ran your automated tests yet.  As soon as I get them working I will post a new PR.

AlanCoding:
This was a helpful contribution, and I agree with the core principles of this PR. We borrowed from this, and then in the PR #68 expanded and integrated it in a way that we believe will be the most consistent with Ansible and other user's requests. This issue should be taken care of and will be put in the next release, so I will close this PR. Thanks.

cedub:
Excellent.  Thank you!

Issue #48, opened 2015-06-09
default perms on ~/.tower_cli.cfg


Hefeweizen:
By default ~/.tower_cli.cfg is created in a manner that relies on system umask.  For any other file, I'd say that was reasonable, but as this will have a plaintext password in it, I'd advocate for the more restrictive 0600 permissions.

mikeycmccarthy:
+1

PR #57, opened 2015-08-24
.tower_cli.cfg permissions set to read/write for current user only

AlanCoding:
Previously, the configuration file was created and edited with the default umask. This commit changes that behavior so that it attempts to set the file permissions to "-rw-------" every time. That is, read/write permissions for current user, no permissions for group, and no permissions for others. The [octal numeric notation](https://en.wikipedia.org/wiki/File_system_permissions#Numeric_notation) is 0600 for this permission.

This is desirable when the configuration file is stored in ~/.tower_cli.cfg, which is in the user's directory, as pointed out in [Issue #48](https://github.com/ansible/tower-cli/issues/48). If the flag --scope=global --scope=local is set, this will attempt to do the same thing, but the ideal behavior for those cases is less clear. Tests were added for the default scope setting and for scope=local.

One other test using warn.assert_called_once was edited to include the warning's arguments with warn.assert_called_once_with, since official documentation for the former can't be found anywhere.

### DIFF
--- a/lib/tower_cli/commands/config.py
+++ b/lib/tower_cli/commands/config.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 import os
+import stat
 import warnings
 
 import six
@@ -147,11 +148,11 @@ def config(key=None, value=None, scope='user', global_=False, unset=False):
     with open(filename, 'w') as config_file:
         parser.write(config_file)
     try:
-        os.chmod(filename, int('0600', 8)) # give rw permissions to user only
+        os.chmod(filename, stat.S_IRUSR | stat.S_IWUSR) # give rw permissions to user only
                                            # fix for issue number 48
     except Exception as e:
         warnings.warn('Unable to set permissions on {0} - {1} '.format(filename,e),
-                            UserWarning)
+                      UserWarning)
     click.echo('Configuration updated successfully.')
 
 
--- a/lib/tower_cli/commands/config.py
+++ b/lib/tower_cli/commands/config.py
@@ -146,6 +146,12 @@ def config(key=None, value=None, scope='user', global_=False, unset=False):
         parser.set('general', key, value)
     with open(filename, 'w') as config_file:
         parser.write(config_file)
+    try:
+        os.chmod(filename, int('0600', 8)) # give rw permissions to user only
+                                           # fix for issue number 48
+    except Exception as e:
+        warnings.warn('Unable to set permissions on {0} - {1} '.format(filename,e),
+                            UserWarning)
     click.echo('Configuration updated successfully.')
 
 
--- a/lib/tower_cli/conf.py
+++ b/lib/tower_cli/conf.py
@@ -16,13 +16,12 @@
 import contextlib
 import copy
 import os
+import stat
 import warnings
 
 import six
 from six.moves import configparser
 from six import StringIO
-
-from sdict import adict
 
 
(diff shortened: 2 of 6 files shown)


lib/tower_cli/commands/config.py
```diff
@@ -146,6 +146,13 @@ def config(key=None, value=None, scope='user', global_=False, unset=False):
         parser.set('general', key, value)
     with open(filename, 'w') as config_file:
         parser.write(config_file)
+        try:
+            os.chmod(filename, int('0600', 8)) # give rw permissions to user only
+                                               # fix for issue number 48
+        except:
+            warnings.warn('Config file permissions can not be edited, which means '
+                          'you could expose your login credentials to other users.',
+                          UserWarning)
```
jlaska: I'd probably simplify this warning message as follow ...

```
Unable to set file permissions on <filename> - <errstr>
```

I'd probably add a call to `warnings.warn()` whenever reading the config file, and `os.stat` reports that the file is group || world readable.

lib/tower_cli/conf.py
```diff
@@ -37,6 +37,16 @@ def _read(self, fp, fpname):
         """
         # Attempt to read the file using the superclass implementation.
         #
+        # Check the permissions of the file we are considering reading
+        # if the file exists and the permissions expose it to reads from
+        # other users, raise a warning
+        if os.path.isfile(fpname):
+            file_permission = os.stat(fpname)
+            nth_permission = [(file_permission.st_mode/8**i)%8 for i in range(3)]
```
matburt: It would probably be cleaner/more readable here to explicitly check the masking bits here:

```
if (file_permission.st_mode & os.IRGRP) or (file_permission.st_mode & os.IROTH) ...
```

for example

lib/tower_cli/commands/config.py
```diff
@@ -146,6 +146,12 @@ def config(key=None, value=None, scope='user', global_=False, unset=False):
         parser.set('general', key, value)
     with open(filename, 'w') as config_file:
         parser.write(config_file)
+    try:
+        os.chmod(filename, int('0600', 8)) # give rw permissions to user only
```
matburt: Make it cleaner by ORing the stat bits together that you want:

```
os.chmod(filename, stat.S_IRUSR | stat.S_IWUSR)
```

matburt:
Those properties mentioned as line items above ^^ can be applied throughout this patch to make it a bit more pythonic.

tests/test_conf.py
```diff
@@ -69,3 +69,18 @@ def test_parser_read_without_header(self):
...
+
+    def test_file_permission_warning(self):
+        """Warn file permissions may expose credentials
+        """
+        with mock.patch.object(warnings, 'warn') as warn:
+            with mock.patch.object(os.path, 'isfile') as isfile:
+                with mock.patch.object(os, 'stat') as os_stat:
+                    isfile.return_value = True
+                    import posix
+                    test_statz = posix.stat_result([0o604,0,0,0,0,0,0,0,0,0])
```
matburt: This is pretty gnarly looking, you can probably use the bitmask here too to make it a bit cleaner without using the posix module.

jlaska:
@AlanCoding looks good to me.  @matburt any objections or concerns before merging?

matburt:
Looks good!

jlaska:
Merging, thanks for the contribution!

PR #60, opened 2015-09-01
Resolve #52 with special test for "create" and adjustments to "modify"

### LINKED ISSUE #52: Creating a new project does not assign to organization
When I create a new project using tower-cli I have to specify the organization, yet it seems to be ignored.

```
(pip)rMBP:scripts$ tower-cli project create --organization "Bovine Inc" --name test --scm-type git --scm-url https://github.com/phips/tiad_demo.git
== ==== ======== ====================================== ========== 
id name scm_type                scm_url                 local_path 
== ==== ======== ====================================== ========== 
26 test git      https://github.com/phips/tiad_demo.git _26__test
== ==== ======== ====================================== ========== 
```

screen shot 2015-07-02 at 18 51 54

AlanCoding:
The resource model for projects has two extra method added to it. The first of these, create(), solves the problem as expressed in issue #52. The second of these, modify(), raising a warning giving the user notice that the organization can not be set on a pre-existing project with the modify command.

Tower-CLI can not be easily configured to accept lists for arguments, and this is the reason for the state of modifying projects. However, it is understandable that someone would want an ability to modify a project's organization. In order to cover this, I added associate_project and disassociate_project to any organization's commands. This should serve as a complete functional patch for the issue.

Corresponding tests were also written for the different cases. I am seeing 99% on resource.project right now, which I could return to fix. Otherwise, tests are added to maintain coverage of the new code.

### DIFF
--- a/tests/test_resources_organization.py
+++ b/tests/test_resources_organization.py
@@ -54,7 +54,7 @@ class OrganizationTests(unittest.TestCase):
                              json.dumps({'disassociate': True, 'id': 84}))
 
     def test_project_associate(self):
-        """Test requests for associating project
+        """Test request for associating project with organization.
         """
         with client.test_mode as t:
             t.register_json('/organizations/42/projects/?id=84',
@@ -65,7 +65,7 @@ class OrganizationTests(unittest.TestCase):
                              json.dumps({'associate': True, 'id': 84}))
 
     def test_project_disassociate(self):
-        """Test requests for associating project
+        """Test request for disassociating project with organization.
         """
         with client.test_mode as t:
             t.register_json('/organizations/42/projects/?id=84',
--- a/tests/test_resources_organization.py
+++ b/tests/test_resources_organization.py
@@ -52,3 +52,26 @@ class OrganizationTests(unittest.TestCase):
             self.org_resource.disassociate(42, 84)
             self.assertEqual(t.requests[1].body,
                              json.dumps({'disassociate': True, 'id': 84}))
+
+    def test_project_associate(self):
+        """Test requests for associating project
+        """
+        with client.test_mode as t:
+            t.register_json('/organizations/42/projects/?id=84',
+                            {'count': 0, 'results': []})
+            t.register_json('/organizations/42/projects/', {}, method='POST')
+            self.org_resource.associate_project(42, 84)
+            self.assertEqual(t.requests[1].body,
+                             json.dumps({'associate': True, 'id': 84}))
+
+    def test_project_disassociate(self):
+        """Test requests for associating project
+        """
+        with client.test_mode as t:
+            t.register_json('/organizations/42/projects/?id=84',
+                            {'count': 1, 'results': [{'id': 84}],
+                             'next': None, 'previous': None})
+            t.register_json('/organizations/42/projects/', {}, method='POST')
+            self.org_resource.disassociate_project(42, 84)
+            self.assertEqual(t.requests[1].body,
+                             json.dumps({'disassociate': True, 'id': 84}))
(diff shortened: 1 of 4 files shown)


tests/test_resources_organization.py
```diff
@@ -52,3 +52,26 @@ def test_disassociate(self):
...
+        with client.test_mode as t:
+            t.register_json('/organizations/42/projects/?id=84',
+                            {'count': 0, 'results': []})
+            t.register_json('/organizations/42/projects/', {}, method='POST')
+            self.org_resource.associate_project(42, 84)
+            self.assertEqual(t.requests[1].body,
+                             json.dumps({'associate': True, 'id': 84}))
+
+    def test_project_disassociate(self):
+        """Test requests for associating project
```
jlaska: `Test requests for s/associating/disassociating/ a project`

tests/test_resources_project.py
```diff
@@ -35,6 +36,62 @@ class UpdateTests(unittest.TestCase):
...
+            t.register_json(endpoint, {'changed': True, 'id': 42},
+                            method='POST')
+            self.res.create(name='bar', scm_type="git")
+            self.assertEqual(t.requests[0].method, 'GET')
+            self.assertEqual(t.requests[1].method, 'POST')
+            self.assertEqual(len(t.requests), 2)
+
+    def test_modify_organization_warning(self):
+        """Test that warning is thrown when trying to modify organization
+        because that can not be done in this way.
```
jlaska: Nit, but `s/can not/cannot/`

tests/test_resources_project.py
```diff
@@ -35,6 +36,62 @@ class UpdateTests(unittest.TestCase):
...
+                                method='GET')
+                t.register_json('/projects/1/', {'name':'bar', 'id':1,
+                                'type':'project', 'organization':1},
+                                method='PATCH')
+                self.res.modify(name='bar', organization=org_pk,
+                                scm_type="git")
+                self.assertEqual(t.requests[0].method, 'GET')
+                self.assertEqual(t.requests[1].method, 'PATCH')
+                self.assertEqual(len(t.requests), 2)
+                warn.assert_called_once_with(mock.ANY, UserWarning)
```
jlaska: Is it important to verify that the error matches the error we expected?

tests/test_resources_project.py
```diff
@@ -35,6 +36,62 @@ class UpdateTests(unittest.TestCase):
...
+                                'type':'project', 'organization':1},
+                                method='GET')
+                t.register_json('/projects/1/', {'name':'bar', 'id':1,
+                                'type':'project', 'organization':1},
+                                method='PATCH')
+                self.res.modify(name='bar', organization=org_pk,
+                                scm_type="git")
+                self.assertEqual(t.requests[0].method, 'GET')
+                self.assertEqual(t.requests[1].method, 'PATCH')
+                self.assertEqual(len(t.requests), 2)
```
jlaska: Above, you are registering 3 json endpoint httpd methods.  Here, you assert that only 2 are called.  Seems like we are have a mismatch?  Are lines 82-84 needed?  Does `tower-cli` issue a `GET` against `/projects/`, followed by a `GET` against `/projects/1/`, followed by a `PATCH` against `/projects/1/`?

jlaska:
@AlanCoding Thanks for the contribution!  Added a  few comments on the proposed changes.  Back to you for review.

tests/test_resources_project.py
AlanCoding: You are correct that I was over-registering the endpoints. Taking out /projects/1/ leaves it completely unaffected.

matburt:
Looks good to me :+1:

jlaska:
Merging, thanks @AlanCoding for the contribution!

PR #62, opened 2015-09-03
features added to resolve #35, but warning on by default

### LINKED ISSUE #35: Problem with SSL: InsecureRequestWarning
I bought a SSL certificate (wildcard domain) to put behind my tower.
When I use `tower-cli` it gives me a warning : 

```
/usr/local/lib/python2.7/dist-packages/requests/packages/urllib3/connectionpool.py:734: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.org/en/latest/security.html
  InsecureRequestWarning)
```

Tower screenshot :

AlanCoding:
The intent is to give user's the option to disable the insecure https warning, if they choose. 2 ways of doing this are added. You can add an --insecure flag to your command and it will suppress the warning. Or, you can set the config variable verify_ssl to false. The method for the config variable was taken from @wybczu in a removed pull request, https://github.com/ansible/tower-cli/pull/29

My hope is that by maintaining the default (and handling the case where the variable doesn't exist) so that it displays the warning, we still respect the intention that people expressed in previous discussions. It would still be preferable to add the proper CA certificate. A test was added to maintain coverage, but we could do better in the future by testing explicitly that the warnings were or were not thrown in a wide range of use cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ $ tower-cli config key value
 By issuing `tower-cli config` with no arguments, you can see a full list
 of configuration options and where they are set.
 
-You will generally need to set at least three configuration options--`host`, 
+You will generally need to set at least three configuration options--`host`,
 `username`, and `password`--which correspond to the location of
 your Ansible Tower instance and your credentials to authenticate to Tower.
 
@@ -81,7 +81,7 @@ $ tower-cli config password myPassw0rd
 
 #### Write to the config files directly.
 
-The configuration file can also be edited directly.  A configuration file is a 
+The configuration file can also be edited directly.  A configuration file is a
 simple file with keys and values, separated by `:` or `=`:
 
 ```yaml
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -166,6 +166,22 @@ $ tower-cli user --help # resource specific help
 $ tower-cli user create --help # command specific help
 ```
 
+#### SSL
+
+By default tower-cli will warn if the SSL certificate of the Tower server
+can not be verified. To disable this warning, set the config variable
```
jlaska: `s/can not/cannot/`

README.md
```diff
@@ -166,6 +166,22 @@ $ tower-cli user --help # resource specific help
...
+`verify_ssl` to true. To disable it just for a single command, add the
+--insecure flag.
+
+```bash
+# Disable insecure connection warnings permanently
+$ tower-cli config verify_ssl true
```
jlaska: The example listed is to *disable* warnings.  Should this be ...

```
$ tower-cli config verify_ssl false
```

jlaska:
@AlanCoding looks good, thanks for the contribution.  A few minor corrections, but should be good to merge afterwards.

jlaska:
Thanks for the fixes @AlanCoding!  Merging.

PR #63, opened 2015-09-03
instances of /etc/awx replaced with /etc/tower, resolves #46

### LINKED ISSUE #46: Latest version of tower does not have /etc/awx
In order to be compatible with the latest ansible tower version the directory /etc/awx does not exist so therefore we need to update tower-cli in order to keep compatibility.

Inside of tower-cli/conf.py the code still looks for /etc/awx for the .tower_cli.cfg for a global config.  This location does not exist for ansible-tower-2.1.3-1.el7.

global_settings = 'tower_cli.cfg' in os.listdir('/etc/awx/')

AlanCoding:
Newer versions of tower use /etc/tower instead of /etc/awx. This is only relevant if --scope=global is set to begin with, and if that folder exists. It is a straightforward change. After this, the older version will no longer be supported.

### DIFF
--- a/lib/tower_cli/commands/config.py
+++ b/lib/tower_cli/commands/config.py
@@ -48,10 +48,10 @@ def config(key=None, value=None, scope='user', global_=False, unset=False):
     """Read or write tower-cli configuration.
 
     `tower config` saves the given setting to the appropriate Tower CLI;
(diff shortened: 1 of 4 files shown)


jlaska:
Looks great, thanks @AlanCoding!

PR #65, opened 2015-09-03
upgrade to resolve #64, more types of credentials

### LINKED ISSUE #64: Support creating all cloud_credential types
`tower-cli` apparently does not support creating VMware, or OpenStack credentials.  This issue is about reviewing current support, and extending `tower-cli` to support all cloud_credential types supported by Ansible Tower.

AlanCoding:
Added 2 options to the list of types of credentials, which were VMware and OpenStack. Adding these also required adding some extra fields, which I believe is also a matter of version changes. More of these types of changes will need to happen to bring tower-cli completely up to speed and resolve some other issues.


jlaska:
Fantastic, thanks for the contribution!

PR #66, opened 2015-09-17
resolve #54 by updating fields, other fixes

### LINKED ISSUE #54: credentials support with privilege escalation *
tower-cli doesn't support latest Ansible Tower field when creating a credential Machie/ssh type regarding 'Privilege Escalation' (sudo/su/etc..) and 'Privilege Escalation Username' .

AlanCoding:
This update allows the use of the become_username, become_password, and other fields that allow privilege escalation by becoming another user through the su command, and others. This has been manually tested to the point of launching new jobs with an su-type privilege escalation.

In addition to that, the @resources.command decorator is added on the project create method overriding the generic resource version. This is a fix for a problem that a previous commit of mine created, which caused tower-cli to crash when creating a job template.

### DIFF
--- a/lib/tower_cli/models/base.py
+++ b/lib/tower_cli/models/base.py
@@ -65,7 +65,7 @@ class ResourceMeta(type):
 
             # If this method has been overwritten from the superclass, copy
             # any click options or arguments from the superclass implementation
-            # down tot the subclass implementation.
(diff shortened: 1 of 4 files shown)


jlaska:
@AlanCoding looks great, thanks for the contribution.  Please merge when ready.

Issue #44, opened 2015-04-07
Passing extra vars overwrites tower extra vars


michaelsharpe:
I am currently utilizing the tower-cli to launch a job from circleCI.  I have an grunt task to create a file with the build number so my instances can pull down a tarball of the code generated as a build artifact.  

When I use this file as an 'extra-vars' option to pass this build number, it overwrites the extra-vars that I have defined ob my job template on the tower.

Is there any way around this, or a better strategy to take to get the build number from circle ci to the tower?

lukesneeringer:
This is defined by the Tower API behavior, rather than tower-cli.
Can you get the extra vars from the job template, then append the ones you need, and pass the whole thing?

tomfotherby:
Hi @michaelsharpe ,

I had the same problem with Tower CLI 2.1.1. I found I could make a small change to `lib/tower_cli/resources/job.py` and change line 80 from:

    data['extra_vars'] = extra_vars

to

    data['extra_vars'] = data['extra_vars']+extra_vars

In Ubuntu 14.04, this is located at `/usr/local/lib/python2.7/dist-packages/tower_cli/resources/job.py`. 

Once this was done, I could invoke `tower-cli` passing in extra vars using:

    $tower-cli job launch --job-template=48 --extra-vars=<(echo "version: 161")

Hopefully with a future release, such a hack won't be needed.

Issue #39, opened 2015-02-25
Ability to interact with survey


perzizzle:
It would be useful to be able to launch a job that has surveys configured for it. Something to the effect of tower-cli job launch --job-template 111 --survey env=production.

In this example, job-template 111 will have a set of extra_vars defined and a survey question that has a list of environments available as radio buttons. I don't want to have to pass in a file with all the extra_vars, I just want to "answer" the survey (or overwrite what env is defined as in the extra_vars)

mtotheikle:
+1

tristanbes:
+1

tomfotherby:
+1 

We've been migrating our scripts so instead of deploying HEAD, we now deploy a particular `build_num`. The build_num is entered using a survey in Ansible Tower when doing a manual deployment. The problem is that our automatic builds no longer work because we need to pass in the build_num via `tower-cli`

I did notice we can pass the variable we need from the survey like this:

    root@3ab86751d116:/# echo "version: 157" > /tmp/extra_vars.yml                                                                                                                                          
    root@3ab86751d116:/# tower-cli job launch --job-template=48 --extra-vars=/tmp/extra_vars.yml

*but* this clobbers the other variables set up via the Tower job template.

AlanCoding:
There's a deeper problem with tower-cli options for surveys - there basically are none. After doing some testing of problems that @simfarm discovered, I realized that even if the job template sets `ask_variables_on_launch=true`, then it will show the existing extra_vars for that template. Survey variables don't get added to the list which is presented to the user. You have to know the survey variable names in advance (basically defeating the point of a survey). There is no way to view the survey questions in tower-cli, and including them would require additional navigation of the job template endpoints.

We could still use the existing structure of a click.edit() text input in a job launch. It would be straightforward to navigate the API endpoints go gather the survey questions and variable defaults. Then leave the questions as comments. I think users would still prefer an editable text area over a linear question/answer structure (although it would be good to get feedback on this). I would rank this as a top priority for a 2.4.0 release.

tomfotherby:
> although it would be good to get feedback on this

In my case, I didn't need to interact with the survey, I just needed to pass some variables I knew were needed by the survey, so v.2.3.0 solves the issue from me.

perzizzle:
[...]

PR #67, opened 2015-09-21
Merge with existing variables (instead of clobber)

tomfotherby:
Currently, if `--extra-vars=filename` is used with `job launch` , any existing `extra-vars` defined in the job template are ignored and only the ones defined in the file are used. This makes it very hard to automate answering surveys via tower-cli (e.g. filling in a variable).

I think merging together the 2 sets of extra-vars is more expected behaviour and allows more use-cases to work.

---

    $tower-cli version
    Ansible Tower 2.2.0
    Tower CLI 2.1.1

    $ansible --version
    ansible 1.9.1

### DIFF
--- a/lib/tower_cli/resources/job.py
+++ b/lib/tower_cli/resources/job.py
@@ -77,7 +77,7 @@ class Resource(models.MonitorableResource):
         if extra_vars:
             if hasattr(extra_vars, 'read'):
                 extra_vars = extra_vars.read()
-            data['extra_vars'] = extra_vars
+            data['extra_vars'] = data['extra_vars']+extra_vars
         elif data.pop('ask_variables_on_launch', False) and not no_input:
             initial = data['extra_vars']
             initial = '\n'.join((


jlaska:
@tomfotherby this is a behavior of the Tower API endpoints, not of `tower-cli`.  Can you confirm which version of Tower you are using?

AlanCoding:
@tomfotherby Thanks for the contribution. This functionality has been on our radar, and I've been working on a fix for it. Using the syntax in this PR of directly referencing data['extra_vars'] will create issues with other common scenarios. The needed behavior will also change with the Tower release cycle. Very soon, I intend to submit another PR which will cover all of these relevant cases.

Your feedback valuable information, and we will make sure to have it merge the variables as you would logically expect.

tomfotherby:
@AlanCoding - Thanks for the info. I'll close this PR then. The minor change I made solves my use case , but I totally agree more things should be considered. (I never used python before so couldn't really contribute more than this hack)

@jlaska - I'm running Ansible Tower 2.2.0, Tower CLI 2.1.1 . The problem I'm trying having is that the Tower job uses a "survey" to ask the user for a particular variable. Since `tower-cli` doesn't have this, I need to pass a extra-var in the `job launch` which effectively provides the answer to the survey. It would work fine except the other variables configuered in Tower get lost. The hack in this PR fixes this issue. See https://github.com/ansible/tower-cli/issues/39#issuecomment-141939811 for more info.

jlaska:
@tomfotherby thanks for the extra details.  Sounds like we're all working in the same space.  A upcoming PR on tower-cli from @AlanCoding should help with this.  In addition, an upcoming Tower release will alter the handling of extra_vars in the manner you'd expect.  More detail on that later.

tomfotherby:
Sorry, this is a bit late in the game, but I just noticed a older issue that is a duplicate of this one: #44.

jlaska:
@tomfotherby thanks for the issue link.  Yup, that looks awfully similar.  @AlanCoding when your PR lands, can you include a comment that closes #44 ?

jlaska:
Might as well close #43 with your new PR also

PR #68, opened 2015-09-22
Updated features for extra_vars passing

AlanCoding:
These changes add a number features related to passing additional parameters for a job. Those are the ability to:

* Specify arguments on the command line, instead of needing to create a text file
* Consistency with Ansible's syntax of specifying a file by starting with and "@" character
* Add your parameters while still preserving those in Ansible Tower's job template
   * This is not needed as of Ansible Tower 2.3.0, and CLI will act accordingly
* Use any number of --extra_vars flags, allowing you to combine command line input with file contents

We also have tests for all this, as well a few unrelated syntax and test edits to bump up coverage to 100% again after it subtly fell after some other recent commits. The core method remains mostly the same as the pull request #43 by @cedub, which is to parse all the available sources of variables, combine them, then dump them into a single variable definition string that can be passed to the API. So thanks for that contribution.

Usage follows the following pattern for launching a job:

> $ tower-cli job launch --job-template=1 --extra-vars="a: 5" --extra-vars="@filename.yml"

This command combines the variables in the file called filename.yml with the command-line set of variables that just says a=5. The job template can also be created with the same syntax:

> $ tower-cli job_template create --name=test_job_template --project=1 --inventory=1 --playbook=helloworld.yml --machine-credential=1 --extra-vars="a: 5" --extra-vars="@filename.yml"

Note that it combines the sources of extra_vars and stores it as only one yaml string. This means that the modify command works as it always has, but you will only have the ability to enter one --extra-vars flag in that case.


jlaska:
@AlanCoding Ansible supports the `<key>=<val>` syntax (e.g. `--extra_vars foo=bar`).  Will `tower-cli` honor that as well?

AlanCoding:
No, it does not support that. To get that level of correspondence with Ansible, my suggested solution would be import Ansible itself. I will look into that.

perzizzle:
I haven't tried running this but can you speak to how variables will be handled if they are defined in multiple places? Eg if a variable is defined at the commandline and in extra_vars of the job template will they merged or will one side "win?"

AlanCoding:
@perzizzle Yes, this will be covered in the updated README as well. The different sources will be merged if there is more than 1. You can use the flag multiple times, and the final one will take priority. All of them take priority over the job template extra-vars. So your command could include --extra-vars="a=1 b=2" --extra-vars="a=3", and you would end up with a=3 and b=2. You can include a file as in --extra-vars="@filename.yml", but it's priority will be set by its order just like all the other flags.

What matters most is that the job template variables are over-written, and that they're not forgotten when even more variables are passed. It will do that correctly.

jlaska:
@AlanCoding :+1: from me. Nice work!

PR #69, opened 2015-09-25
Flake8

### LINKED ISSUE #58: Enable flake8 tests and include in travis-ci configuration

AlanCoding:
Some features to resolve #58, flake8 and travis integration. One flake8 error still pops up, that 'file' is an undefined name. This syntax is probably antiquated somewhat, and it's difficult to understand what it should be, without breaking something else. For now, it has the --exit-zero flag set, indicating that passing the tests are not mandatory, but this is something up for discussion.

This is not merged with the other pull request. They both branch of master and will need to be merged later.

### DIFF
--- a/tests/compat.py
+++ b/tests/compat.py
@@ -20,21 +20,17 @@
 # Import unittest2 if we have it (required on Python 2.6),
 # otherwise just use the stdlib unittest.
 try:
-    import unittest2 as unittest
+    import unittest2 as unittest  # NOQA
 except ImportError:  # Python >= 2.7
-    import unittest
+    import unittest  # NOQA
 
 try:
-    from unittest import mock
+    from unittest import mock  # NOQA
 except ImportError:  # Python < 3.3
-    import mock
+    import mock  # NOQA
 
 # Import the correct class used for when a file is opened.
 try:
     from io import TextIOWrapper
 except ImportError:  # Python < 3
-    TextIOWrapper = file
-
-
-unittest
-mock
+    TextIOWrapper = file  # NOQA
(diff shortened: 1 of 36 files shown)


tests/compat.py
```diff
@@ -34,3 +34,7 @@
     from io import TextIOWrapper
 except ImportError:  # Python < 3
     TextIOWrapper = file
```
AlanCoding: @jlaska This is the one line that I couldn't could to pass flake8. Here, 'file' is a keyword native to Python... but only for older versions. So I guess it's not surprising that flake8 does not see it as such. But then we're in a bit of a catch 22. How do we keep it functioning correctly and also pass flake8?

jlaska: `# NOQA` ?

AlanCoding: @jlaska Yep, noqa was what I was grasping for. Edited things to use that approach with recent commit. I'll give it a more thorough review later, but that plugs all the major holes.

AlanCoding:
Coveralls seem to be complaining about a decreased coverage percent. I'm not sure, but I think this is because its baseline is the other branch (which does have higher coverage). Either way, I anticipate that will be fixed with a merge.

jlaska:
Bonus, well done @AlanCoding .  The only comment I have regarding the use of `noqa` is that I'd probably prefer seeing that used on a per-line basis, rather than a per-file basis.  If you're okay with that, I'd proceed with merging afterwards.

jlaska:
@AlanCoding looks great, go ahead and merge whenever ready.  Thanks for your contributions :)

PR #70, opened 2015-09-28
add skip-tags to JT resource, fix parser bug

nitzmahone:
recent parser change assumed extra-vars was always specified, bombs when it wasn't

cc @AlanCoding

### DIFF
--- a/lib/tower_cli/resources/job_template.py
+++ b/lib/tower_cli/resources/job_template.py
@@ -56,6 +56,7 @@ class Resource(models.Resource):
         required=False,
     )
     job_tags = models.Field(required=False, display=False)
+    skip_tags = models.Field(required=False, display=False)
     extra_vars = models.Field(required=False, display=False)
     become_enabled = models.Field(type=bool, required=False, display=False,
                                   show_default=True, default=False)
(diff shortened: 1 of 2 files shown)


jlaska:
:+1: looks good, thanks @nitzmahone

AlanCoding:
This is very helpful. It did not occur to me that the key could be in the dictionary but still be None, although it is a possibility. Being that it's an error likely to occur in job_template actions as opposed to job launch actions, that can slip by me.

I will have another look tomorrow morning, and see about writing a corresponding test.

nitzmahone:
@AlanCoding - the case I hit on this was modifying an existing JT to add skip_tags (since as of 2.3, the UI still doesn't support it), so the only values specified were the JT name and the skip_tags value. Should be pretty easy to write a test for that case. Hit me up on Slack (mdavis) if that doesn't make sense...

AlanCoding:
Good that we have the skip_tags as a matter of keeping up to date.

In terms of bombing when a source of extra-vars is None or empty, we still have a problem even if we avoid bombing in the conditional before loading from a file. It gets passed on to the parser, in which case it just bombs in a different way. See 2a5984401fe9740bd67106241dc61be0696527ac to see what I think we should do to deal with those scenarios.

PR #71, opened 2015-09-29
handle more edge cases, update tests

AlanCoding:
Tests were added to broadly cover null values and empty strings in the parser functions. It turned out that the current parser utility would not pass those, so more changes were made in order to have them all passing. @nitzmahone, if you wouldn't mind, could you take a look at what was done here and see if anything else needs to be changed?


jlaska:
Thanks for the contribution!  Looks good :+1:

PR #72, opened 2015-09-29
ad hoc command added to resolve #61, inheritance restructure, tests

### LINKED ISSUE #61: Support Tower 2.2's remote command execution
tower-cli should support Tower 2.2's remote command execution feature. See the https://tower.local/api/v1/ad_hoc_commands/ endpoint.

AlanCoding:
This has the basic gears necessary to satisfy @wenottingham feature request to support ad-hoc commands in modern versions of tower.

In this PR, I have also removed the "create" option from the list of commands on a job. The same goes for ad-hoc commands. They can be launched, but not created. To do this, the models that @lukesneeringer built have been reorganized to reflect the modern demands for different types of resources. This structure is illustrated as follows:

_Classes_
- RM - ResourceMeta
- BR - BaseResource
- MR - MonitorableResource
- RR - ReadableResource
- WR - WritableResource

_Inheritance Structure_
```
 RM --> BR --> RR --> WR
                \--> MR
```
This is just to say, MonitorableResource now branches from BaseResource. I thought this was a minimalist revision since it clearly did not need anything from the super classes. With this, we can use MR for things like job and ad-hoc commands which should not have a create or modify command. Likewise, we can use MR+WR for things like a project which need all of the tools in the toolbox. The prior structure was more linear.

I still have a tiny test gap and probably a lot of other parts which could use improvement, but the current state seems to be working and I would like to solicit comments and code review since I have encroached further toward the core of tower-cli than ever before.

UPDATE: MR was changed to inherit from RR, because it was never used except for in conjunction with RR, and because some MR methods relied on lookup, get, and other reading methods.

### DIFF
--- a/lib/tower_cli/models/base.py
+++ b/lib/tower_cli/models/base.py
@@ -321,6 +321,7 @@ class BaseResource(six.with_metaclass(ResourceMeta)):
 
                 # Sanity check: If there is a "changed" key in our payload
                 # and little else, we print a short message and not a table.
+                # this specifically applies to deletion
                 if 'changed' in payload and 'id' not in payload:
                     return 'OK. (changed: {0})'.format(
                         six.text_type(payload['changed']).lower(),
@@ -329,6 +330,8 @@ class BaseResource(six.with_metaclass(ResourceMeta)):
                 # Sanity check: If there is no ID and no results, then this
                 # is unusual output; keep our table formatting, but plow
                 # over the columns-as-keys stuff above.
+                # this originally applied to launch/status/update methods
+                # but it may become deprecated
                 if 'id' not in payload and 'results' not in payload:
                     columns = [i for i in payload.keys()]
 
@@ -834,7 +837,7 @@ class MonitorableResource(ReadableResource):
                   help='If provided, this command (not the job) will time out '
                        'after the given number of seconds.')
     def monitor(self, pk, min_interval=1, max_interval=30,
-                timeout=None, outfile=sys.stdout):
+                timeout=None, outfile=sys.stdout, **kwargs):
         """Monitor a running job.
 
         Blocks further input until the job completes (whether successfully or
@@ -853,7 +856,7 @@ class MonitorableResource(ReadableResource):
         # and run in Python.  This seems fine; outfile can be set to /dev/null
         # and very much the normal use for this method should be CLI
         # monitoring.
-        result = self.status(pk)
+        result = self.status(pk, detail=True)
         last_poll = time.time()
         timeout_check = 0
         while result['status'] != 'successful':
@@ -905,7 +908,7 @@ class MonitorableResource(ReadableResource):
             # to the next time that we intend to do a check, and once that
             # time hits, we do the status check as part of the normal cycle.
             if time.time() - last_poll > interval:
-                result = self.status(pk)
+                result = self.status(pk, detail=True)
                 last_poll = time.time()
                 interval = min(interval * 1.5, max_interval)
 
@@ -926,6 +929,9 @@ class MonitorableResource(ReadableResource):
             ('id', pk),
         ))
         answer.update(result)
+        # Make sure to return ID of resource and not update number
+        # relevant for project creation and update
+        answer['id'] = pk
         return answer
 
 
@@ -974,9 +980,9 @@ class ExeResource(MonitorableResource):
         Fails with a non-zero exit status if the job cannot be canceled.
         You must provide either a pk or parameters in the job's identity.
         """
-        # Search for the record within its identity if pk not given
+        # Search for the record if pk not given
         if not pk:
-            existing_data = self._lookup(fail_on_missing=True, **kwargs)
+            existing_data = self.get(**kwargs)
             pk = existing_data['id']
 
         cancel_endpoint = '%s%d/cancel/' % (self.endpoint, pk)
--- a/lib/tower_cli/models/base.py
+++ b/lib/tower_cli/models/base.py
@@ -493,6 +493,63 @@ class ReadableResource(BaseResource):
         # Return the response.
         return resp
 
+    # Convenience wrappers around `read` and `write`:
+    #   - read:  get, list
+    #   - write: create, modify
+
+    @resources.command(ignore_defaults=True)
+    def get(self, pk=None, **kwargs):
+        """Return one and exactly one object.
+
+        Lookups may be through a primary key, specified as a positional
+        argument, and/or through filters specified through keyword arguments.
+
+        If the number of results does not equal one, raise an exception.
+        """
+        if kwargs.pop('include_debug_header', True):
+            debug.log('Getting the record.', header='details')
+        response = self.read(pk=pk, fail_on_no_results=True,
+                             fail_on_multiple_results=True, **kwargs)
+        return response['results'][0]
+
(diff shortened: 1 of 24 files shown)


AlanCoding:
With some feedback @jlaska, I have a list of additional improvements to make to this PR before it is merged. Those are:
- Tighten up for the DRY principle, moving cancel and status to a new super class
- Copy/edit the documentation and keep wording (like "command" vs "job") consistent
- Remove default values for many fields and kick off the None-valued fields automatically
- Add more fields to the display tables (list and get)

Should have those changes committed today without any problem.

AlanCoding:
Update:

I don't think the coveralls failure is a big deal. Code was moved out of one file and into another, and the failure amounts to saying that fewer lines are now covered in the first file. Of course it does.

This is now also addressing issue #73, but I must and volunteer the biggest bug with my current solution.

- `$ tower-cli job status --elapsed="4.063"` Object can not be found
- `$ tower-cli job get --elapsed="4.063"` Object can be found

I know this behavior has to do with the filtering that goes on in a get request. The most simple solution might just be to exclude the fields from status and cancel methods. Like launch, these are much more custom than functions on many other resources and that leaves holes like this one.

lib/tower_cli/resources/ad_hoc.py
```diff
@@ -0,0 +1,114 @@
+# Copyright 2015, Ansible, Inc.
+# Luke Sneeringer <lsneeringer@ansible.com>
```
jlaska: You should take credit for this :)

lib/tower_cli/resources/job_template.py
```diff
@@ -58,8 +58,7 @@ class Resource(models.Resource):
     job_tags = models.Field(required=False, display=False)
     skip_tags = models.Field(required=False, display=False)
     extra_vars = models.Field(required=False, display=False)
-    become_enabled = models.Field(type=bool, required=False, display=False,
-                                  show_default=True, default=False)
```
jlaska: Was removing `show_default=True` intentional?

tests/test_resources_ad_hoc.py
```diff
@@ -0,0 +1,143 @@
+# Copyright 2015, Ansible, Inc.
+# Luke Sneeringer <lsneeringer@ansible.com>
```
jlaska: Take some credit! :)

lib/tower_cli/resources/ad_hoc.py
```diff
@@ -0,0 +1,114 @@
...
+
+    # Parameters similar to job_template
+    job_type = models.Field(
+        default='run',
+        display=False,
+        show_default=True,
+        type=click.Choice(['run', 'check']),
+    )
+    inventory = models.Field(type=types.Related('inventory'))
+    machine_credential = models.Field(
```
jlaska: This appears in the `--help` as `--machine_credential`.  Has it always been this way?  I'm lazy, and really tempted to use `--credential`.  Does `click` offer a way to have aliases for a field?

jlaska:
When testing ad_hoc command launch, I'm observing the following failure?  Is this my doing?

```bash
# PYTHONPATH=lib ./bin/tower-cli ad_hoc launch --inventory Local --machine-credential 19 --module-args 'uptime' 
```
```python
  File "./bin/tower-cli", line 80, in <module>
    cli()
  File "/Users/jlaska/Downloads/VirtualEnvs/tower-cli/lib/python2.7/site-packages/click/core.py", line 700, in __call__
    return self.main(*args, **kwargs)
  File "/Users/jlaska/Downloads/VirtualEnvs/tower-cli/lib/python2.7/site-packages/click/core.py", line 680, in main
    rv = self.invoke(ctx)
  File "/Users/jlaska/Downloads/VirtualEnvs/tower-cli/lib/python2.7/site-packages/click/core.py", line 1027, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/jlaska/Downloads/VirtualEnvs/tower-cli/lib/python2.7/site-packages/click/core.py", line 1027, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/jlaska/Downloads/VirtualEnvs/tower-cli/lib/python2.7/site-packages/click/core.py", line 873, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/jlaska/Downloads/VirtualEnvs/tower-cli/lib/python2.7/site-packages/click/core.py", line 508, in invoke
    return callback(*args, **kwargs)
  File "/Users/jlaska/Projects/tower-cli.git/lib/tower_cli/utils/decorators.py", line 49, in answer
    return method(*inner_a, **inner_kw)
  File "/Users/jlaska/Projects/tower-cli.git/lib/tower_cli/models/base.py", line 298, in func
    output = format(result)
  File "/Users/jlaska/Projects/tower-cli.git/lib/tower_cli/models/base.py", line 359, in _format_human
    *[len(six.text_type(i[col])) for i in raw_rows]
KeyError: 'name'
```

lib/tower_cli/resources/job_template.py
AlanCoding: Yes, this has to do with how the CLI interacts with the API, and I should explain this. Fields that are not required will generally (reasonably) not use a default. These fields will not even be included when it does a POST. In the JSON data it's not true or false, it's just not there. Not including a field means that we're relying on whatever kind of default the Tower server has set.

lib/tower_cli/resources/ad_hoc.py
AlanCoding: Tower-cli has always been this way, but I'm thinking that maybe Tower renamed machine_credential to credential. Both of these appear in the overall credentials list, but I think the reason we have both fields is so that you can make a job template with _both_ a machine credential and cloud credential. Renaming machine_credential to credential sounds like a correct step to take, but I don't know the history well enough and my theory could be incorrect.

jlaska: We can track any changes to `machine_credential` in a future issue || PR.  But thanks for helping me understand better in the context of tower-cli.

lib/tower_cli/models/base.py
```diff
@@ -828,7 +837,7 @@ def status(self, pk, detail=False):
                   help='If provided, this command (not the job) will time out '
                        'after the given number of seconds.')
     def monitor(self, pk, min_interval=1, max_interval=30,
-                timeout=None, outfile=sys.stdout):
+                timeout=None, outfile=sys.stdout, **kwargs):
```
AlanCoding: Note: Adding kwargs here enables use for resources that have fields declared on them. Even if those fields are not specified, they are passed with other utilities with None values. This is particularly relevant for the ad hoc commands, because there is a greater need for fields (more similar to job template), rather than just command line flags.

lib/tower_cli/resources/job_template.py
```diff
@@ -76,7 +75,7 @@ def create(self, extra_vars=None, *args, **kwargs):
         return super(Resource, self).create(*args, **kwargs)
 
     @resources.command
-    def modify(self, *args, **kwargs):
+    def modify(self, pk=None, *args, **kwargs):
```
AlanCoding: Note: arguments need to be declared like this in order for the click library to accept command line use of the command, like "command 15", and if not, that usage will throw an error. I thought this type of detail might be important to go in a readme for contributing at some point.

PR #75, opened 2015-10-07
Added more "inventory source" variables on group create / modify

AlanCoding:
Added fields to the (hidden) resource inventory_source and options to the resource group. You can use the group create or group modify command to set these options. Those options include:
- source_regions
- source_vars
- overwrite
- overwrite_vars
- update_on_launch
- update_cache_timeout

Even with these added, not all possibilities are covered. For instance, for EC2 specifically we are not covering:
- instance fileters
- only group by

There are even more if you want to dig in the API, but unless someone specifically articulates a need for these, I'm not going to assume that they should be included at this point.

This survives manual testing. Since there is no logic added, there are no unit tests to add.

Note that if you create a group with the CLI and specify the region, it will not show up in the UI. However, if I specify a region in the UI and save it, it doesn't show in that case either. That might be my own server version, or it might be a UI bug.

I first set this PR to compare the ad_hoc branch as a base. That did not work, so I'm changing it to master in the main repo, and we will go from there.

### DIFF
--- a/lib/tower_cli/models/__init__.py
+++ b/lib/tower_cli/models/__init__.py
@@ -18,6 +18,7 @@
 from __future__ import absolute_import, unicode_literals
 
 from tower_cli.models.base import BaseResource, ReadableResource, \
-                                  WritableResource, MonitorableResource
+                                  WritableResource, MonitorableResource, \
+                                  ExeResource
 from tower_cli.models.fields import Field
 from tower_cli.utils.types import File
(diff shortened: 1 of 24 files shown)


lib/tower_cli/models/__init__.py
```diff
@@ -12,9 +12,13 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+#
+# flake8: noqa
```
jlaska: I tend to avoid module-level exclusions, but it seems to make sense for this case.

PR #76, opened 2015-10-08
ad hoc command added to resolve #61, column edit to resolve #73

### LINKED ISSUE #61: Support Tower 2.2's remote command execution

### LINKED ISSUE #73: Add more detail to job `list` and `get` output
Currently, the output from `tower-cli job list` and `tower-cli job get` is too terse and only includes an `id`.  Perhaps additional attributes could be added, such as `name`, `inventory`, `launched_by` etc...

This issue is about deciding which fields are relevant to surface, and adding them to the `tower-cli job [list|get]` sub-commands.

AlanCoding:
Ad hoc command added and some changes were correspondingly made to job, as these are both implemented through the ExeResource class (an extension of MonitorableResource), covering status and cancel. The previous class Resource in base.py was renamed ResouceMethods but retains the existing ordering of methods in the master branch, and this was changed from the last version of this PR. The "create" method is converted into a command inside of the Resource class in base.py so that the majority of resources can continue to inherit from it in the same way as before, but still leaving off the method from the command-line options for the cases of job and ad hoc resources.

Several changes were also made in order to print a free-form list of columns when running, monitoring, or checking the status of an ad hoc command, but these changes also apply to running a job. These methods were excluded from the special case in the _format_human method that checked if a unique combination of keys existed in the returned dictionary, and narrowed the list of columns if that were the case. Almost every other resource+command combination (which one set of exceptions, I can elaborate on later) has the fields flagged with display=True printed. Since ad hoc commands borrow fields (which are typically displayed in this way) from job template, it was written to follow this format, even though doing so demanded making some changes to the way status, cancel, and monitor operate. That is what is typically going on when you see **kwargs addition in the code, and I have attempted to be as meticulous as possible checking to see that these alterations don't negatively affect anything else.

Additionally, some errors with lookups were changed in a way that affected more resources than just ad hoc. For an example of what I'm talking about, I would like to be able to call `job launch --job-template=hello_world` as well as `job launch --job-template=7`, if that specific job template has a pk of 7. Switching logic for integers versus names should be pretty automatic, and it typically is, but some cases where the modify method is over-ridden (for instance), and when multiple contributors made changes to these implementations, support for one or the other has been unintentionally dropped. Some edits here deal with that, either with the pk specifier, or a custom set of logic that has already been used in a number of resources here and there. A more thorough audit of this behavior is left for the future.

One problem with test coverage did come up for this PR in particular. There is a statement that replaces all occurrences of underscores with hyphens in the help_text. I don't actually see where underscores were ever used in the help text, so this explains why it was not covered. It's harder to explain why it was covered in the past, but I will look into it.

### DIFF
--- a/lib/tower_cli/models/base.py
+++ b/lib/tower_cli/models/base.py
@@ -24,6 +24,7 @@ import re
 import sys
 import time
 from copy import copy
+from sdict import adict
 
 import six
 
@@ -320,6 +321,7 @@ class BaseResource(six.with_metaclass(ResourceMeta)):
 
                 # Sanity check: If there is a "changed" key in our payload
                 # and little else, we print a short message and not a table.
+                # this specifically applies to deletion
                 if 'changed' in payload and 'id' not in payload:
                     return 'OK. (changed: {0})'.format(
                         six.text_type(payload['changed']).lower(),
@@ -328,6 +330,8 @@ class BaseResource(six.with_metaclass(ResourceMeta)):
                 # Sanity check: If there is no ID and no results, then this
                 # is unusual output; keep our table formatting, but plow
                 # over the columns-as-keys stuff above.
+                # this originally applied to launch/status/update methods
+                # but it may become deprecated
                 if 'id' not in payload and 'results' not in payload:
                     columns = [i for i in payload.keys()]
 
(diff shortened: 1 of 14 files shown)


jlaska:
:+1:

lib/tower_cli/models/base.py
```diff
@@ -914,5 +919,99 @@ def monitor(self, pk, min_interval=1, max_interval=30,
...
+    @click.option('--fail-on-found', default=False,
+                  show_default=True, type=bool, is_flag=True,
+                  help='If used, return an error if a matching record already '
+                       'exists.')
+    @click.option('--force-on-exists', default=False,
+                  show_default=True, type=bool, is_flag=True,
+                  help='If used, if a match is found on unique fields, other '
+                       'fields will be updated to the provided values. If '
+                       'False, a match causes the request to be a no-op.')
+    def create(self, fail_on_found=False, force_on_exists=False, **kwargs):
```
AlanCoding: Found it - this is the part that caused the coverage decline. doc strings on commands are more than just simple suggestions. They are used in the help text on the command line, so the descriptive text should have been moved here. The parsing function was intended to operate on a specific type of string in this help text (which apparently doesn't exist in other, similar, methods), and it doesn't exist because of inadvertent over-riding.

jlaska: @AlanCoding good catch!

PR #78, opened 2015-10-08
Bash script containing examples of tower-cli use

AlanCoding:
This bash script generates fake data. Several things are necessary in order to use it:
- Config variables are set to a valid host, username, and password
- You have fake user accounts user1 and user2, as described in the docs

There is also a script to attempt to remote the same data that was created.

It is hoped that this will be a good addition to the examples in the README file, giving much more complete examples of using tower-cli.


jlaska:
Looks good, thanks!  Please merge at your convenience.

PR #79, opened 2015-10-08
Additional inventory source options added

AlanCoding:
Additional fields have been added to the "group create" and "group modify" commands. These are passed into the hidden resource inventory_source and from there the API is updated. This solution has been tested manually and there is an example in the script example in another branch.

Additional options include
- source_regions
- source_vars
- overwrite
- overwrite_vars
- update_on_launch
- update_cache_timeout

Even with these added, not all possibilities are covered. For instance, for EC2 specifically we are not covering:

- instance fileters
- only group by

Since there is no logic added, there are no unit tests to add.

There were 2 failed PR filed for this issue, but they had problems with the source it branched off of, and this version should work for the new release.

### DIFF
--- a/lib/tower_cli/resources/group.py
+++ b/lib/tower_cli/resources/group.py
@@ -50,7 +50,7 @@ class Resource(models.Resource):
     @click.option('--overwrite-vars', type=bool,
                   help='Override vars in child groups and hosts with those '
                   'from the external source.')
-    @click.option('--update-on-launch', type=bool, help='Refersh inventory '
+    @click.option('--update-on-launch', type=bool, help='Refresh inventory '
                   'data from its source each time a job is run.')
     def create(self, credential=None, source=None, **kwargs):
         """Create a group and, if necessary, modify the inventory source within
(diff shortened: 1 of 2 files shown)


jlaska:
@AlanCoding what's up with the coverage failure?

lib/tower_cli/resources/group.py
```diff
@@ -75,6 +87,17 @@ def create(self, credential=None, source=None, **kwargs):
...
+    @click.option('--source-regions', help='Regions for your cloud provider.')
+    # Options may not be valid for certain types of cloud servers
+    @click.option('--source-vars', help='Override variables found on source '
+                  'with variables defined in this field.')
+    @click.option('--overwrite', type=bool,
+                  help='Delete child groups and hosts not found in source.')
+    @click.option('--overwrite-vars', type=bool,
+                  help='Override vars in child groups and hosts with those '
+                  'from the external source.')
+    @click.option('--update-on-launch', type=bool, help='Refersh inventory '
```
jlaska: Typo ... `Refresh`

jlaska:
@AlanCoding other than the typo, looks good.  Go ahead and merge after fixing.

PR #80, opened 2015-10-09
editing readmes in tower cli for active voice and clarity

docschick:


AlanCoding:
Looks good :thumbsup:

PR #81, opened 2015-10-09
Veryify_ssl config var changed from a warning message flip to an error flip

AlanCoding:
Assume we are dealing with a server of Tower that does not have a SSL certificate found in any of the computers certificate authorities.

- Past behavior: if verify_ssl was true, you would get the standard insecure request warning message
- New behavior: if verrify_ssl is true, the connection will be rejected and you will receive an error

The new error message for this scenario is the following.

> Error: Could not establish a secure connection. Please add the server to your certificate authority.
> You can run this command without verifying SSL with the --insecure flag, or permanently disable verification by the config setting:
> 
>  tower-cli config verify_ssl false

If the host is not available, you should still see the same message as always. Care was taken to not conflate those scenarios.

Also, the release notes for the coming 2.3.0 were added in the HISTORY.md file, top level directory.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,0 +1,24 @@
+## Release History
+
+## 2.0.0 (2014-07-10)
+
+* Pluggable resource architecture built around click
+
+## 2.0.1 (2014-07-18)
+
+## 2.0.2 (2014-10-02)
+
+### 2.1.0 (2015-01-21)
+
+### 2.3.0 (2015-10-xx)
+
+* Fixed an issue with the settings file so that it could be world readable
+* Added the ability to associate a project with an organization
+* Added setting "verrify_ssl" to disallow insecure connections
+* Added support for additional cloud credentials
+* Exposed additional options for a cloud inventory source
+* Combined " launch-time extra_vars" with " job_template extra_vars" for older Tower versions
+* Changed the extra_vars parameters to align with Ansible parameter handling
+* Added the ability to run ad hoc commands
+* Included more detail when displaying job information
+* Added an example bash script to demonstrate tower-cli usage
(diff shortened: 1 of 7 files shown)


HISTORY.md
```diff
@@ -0,0 +1,24 @@
...
+
+## 2.0.1 (2014-07-18)
+
+## 2.0.2 (2014-10-02)
+
+### 2.1.0 (2015-01-21)
+
+### 2.3.0 (2015-10-xx)
+
+* Fixed an issue with the settings file so that it could be world readable
```
AlanCoding: * so that it will _not_ be world-readable

lib/tower_cli/api.py
```diff
@@ -88,16 +88,34 @@ def request(self, method, url, *args, **kwargs):
...
-            urllib3.disable_warnings()
+            verify_ssl = False
 
         # Call the superclass method.
         try:
             with warnings.catch_warnings():
-                r = super(Client, self).request(method, url, *args,
-                                                verify=False, **kwargs)
+                warnings.simplefilter(
+                    "ignore", urllib3.exceptions.InsecureRequestWarning)
```
AlanCoding: The "with" block and with the simplefilter function are both necessary in order to suppress the warning. If the user set verify_ssl=false, then they have demonstrated consent to any risks, so they shouldn't want to see additional warnings.

tests/test_api.py
```diff
@@ -155,13 +160,20 @@ def test_bad_request_error(self):
             with self.assertRaises(exc.BadRequest):
                 client.get('/ping/')
 
-    def test_disable_connection_warning(self):
+    def test_insecure_connecction(self):
```
matburt: connecction huh? ;)

matburt:
Other than the spelling of `test_insecure_connection` this looks good to me

PR #82, opened 2015-10-12
README file converted to rst in order to render in PyPI

AlanCoding:
This pull request renames README.md in the top level directory to README.rst. Formatting changes were made as-necessary in order to keep the look and feel of the README in github the exact same, and it appears to display correctly in alancoding/README_to_rst.

For demonstration, see the PyPI page for the most recent tower-cli release. This page does not render the markup on the tower-cli github page.

https://pypi.python.org/pypi/ansible-tower-cli

For demonstration of a counter-example which does render in its PyPI page, see the requests project, which uses an rst README file. They also have a release history appended onto the readme, but I will leave this for future work.

https://pypi.python.org/pypi/requests/

I tried to keep the file history in-tact by using `git mv README.md to README.rst`, but unfortunately, the git history still sees a file deletion and then file creation. As far as I know, this shouldn't break anything else and it would make the PyPI page look more professional.


jlaska:
@AlanCoding nicely done, thanks!

AlanCoding:
As a counter-point, the PyPI page by @jlaska for pytest-ansible somehow managed to get the .md format to render correctly, badges and all.
https://pypi.python.org/pypi/pytest-ansible
If possible, that would be preferable to making this change. For now, I will leave it here and when we go to compile it for release I will look for a specific option to select the markdown type. If that can't be found, we can merge this into master and then go from there.

jlaska:
Look at the setup.py for pytest-ansible. I use pandoc to convert the markdown when building.

PR #83, opened 2015-10-14
Release cleanup, changing docs to .rst

AlanCoding:
Summary of changes:

- Change README to rst, this time using pypandoc (which leaves the syntax more readable)
- File histories still remain intact using the --follow tag, but things like git blame will be forever broken, since there is no way to preserve this when renaming a file
- Expanded HISTORY record
- HISTORY added to long_description in setup.py, inspired by [pytest Ansible](https://github.com/jlaska/pytest-ansible), but not using the exact same approach

This pull request combines methods from several other examples in order to (hopefully) get a PyPI description page looking like we want it. I wanted to avoid importing pypandoc, because it is not installed _just_ with a pip requirement, so I believe we would need an "apt-get" command or something like that. I was fairly certain this would cause problems in tox or some other script for the tower-cli release workflow. As such, we convert to .rst once here and leave it that way. On the other hand, I did borrow the approach for combining files from pytest-ansible and added HISTORY to the manifest, so this will be shown on the PyPI page.

These are minor but important differences from my last PR (just closed). This is mainly giving more consideration to the details for switching the README to rst in this release.

Also added a line in the script to create the Default organization, since that may not be present initially.


AlanCoding:
Exactly, as you said:

> Having it listed in the `requirements.txt` means `pypandoc` will be installed whenever someone does `pip install ansible-tower-cli`.  I don't think that's the desired behavior.

This is the route of reasoning I followed. Putting it in the main `requirements.txt` file would be a bad idea. Say that we did want to implement some automatic conversion process. Ideally, that conversion would only be triggered when we are pushing a release. So, just like we have `tests/requirements.txt`, we could add a 3rd list of requirements. That would raise a lot of integration issues and it would probably be better to avoid that. Best just to leave it out altogether.

> My impression with your use of `pypandoc` was that it was a one-time thing used to convert existing markdown documents. 

Yes, because the output from pypandoc was a lot easier to read than my previous attempt. It seems reasonable to expect future contributors to maintain the README and HISTORY files from this PR. We are trying to strike a balance between that concern, getting the PyPI formatting right, and not breaking anything else. All of those considerations factored into the decisions made here.

jlaska:
Looks good @AlanCoding, please merge at your convenience

PR #84, opened 2015-10-15
Remove egg_base variable (for Windows compatibility) and add to .gitignore

AlanCoding:
In response to pull request #34, we are removing the egg_base variable from setup.cfg, which avoids using an OS-specific directory. Also add this type of file to .gitignore to avoid committing unnecessary files later.


jlaska:
@AlanCoding thanks, looks good!

Issue #56, opened 2015-08-21
tower-cli parameter names should match API attribute names


simfarm:
Some error messages in tower-cli need to get updated when creating a project.

```
Error: Missing required fields: organization, scm_type
```
Needs to get updated. The actual flag is `scm-type`. Also:

```
Error: The Tower server claims it was sent a bad request.

POST https://172.16.239.132/api/v1/projects/
Params: None
Data: {"scm_type": "git", "organization": 1, "name": "foobles"}

Response: {"scm_url": ["SCM URL is required"]}
```

Needs to get updated. The actual flag is `scm-url`.

jlaska:
> Needs to get updated. The actual flag is scm-type. Also:

@simfarm are you certain?  The `/api/v1/projects/` endpoint accepts the following attributes.

```json
            "created": "2015-07-07T17:35:22.970Z", 
            "modified": "2015-07-27T21:09:50.757Z", 
            "name": "ansible/ansible-it", 
            "description": "", 
            "local_path": "_92__meyers_ansible_it", 
            "scm_type": "git", 
            "scm_url": "git@github.com:ansible/ansible.git", 
            "scm_branch": "", 
            "scm_clean": true, 
            "scm_delete_on_update": true, 
            "credential": 2, 
            "last_job_run": "2015-07-28T03:24:05.336Z", 
            "last_job_failed": false, 
            "has_schedules": false, 
            "next_job_run": null, 
            "status": "successful", 
            "scm_delete_on_next_update": false, 
            "scm_update_on_launch": false, 
            "scm_update_cache_timeout": 0, 
            "last_update_failed": false, 
            "last_updated": "2015-07-28T03:24:05.336Z"
```
The attribute `scm_type` *is* correct.  In addition, `scm_url` is correct.

simfarm:
@jlaska Thanks for looking in to this. Check it out, `tower-cli` accepts `scm-type` and `scm-url`, NOT `scm_type` and `scm-url`.

```
[root@localhost ~]# tower-cli project create --name Name --scm_type git
Error: no such option: --scm_type
[root@localhost ~]# tower-cli project create --name Name --scm-type git
[...]

PR #85, opened 2015-10-16
Convert "_" to "-" in tower-cli missing field error message

AlanCoding:
Issue #56 brought up a confusing behavior displayed by tower-cli, where sometimes its language would use underscores in variables names, and other times it would use dashes instead.

We hold to our previous design decision to use dashes, being consistent with Ansible core and general good practice for command-line based programs. However, there are some small things we can do to clean up error messages that confusingly display underscores. There is no changing the fact that underscores and used in the API, but the user of tower-cli should not have to think about this, or be exposed to variables with underscores in them. For this one particular scenario, I suggest that we just show the user the variable name with dashes.

This fixes the error message when you don't give a required field - as defined by tower-cli. However, there are many special case scenarios where a field is only required if another field is given a certain value. In those cases, the user will be shown an error response from the Tower server, and those will contain underscores. Patching up these responses would be more complicated because you would have to parse it as JSON first. I will leave that work for the future.

### DIFF
--- a/lib/tower_cli/models/base.py
+++ b/lib/tower_cli/models/base.py
@@ -467,7 +467,8 @@ class ResourceMethods(BaseResource):
         # If queries were provided, process them.
         for query in queries:
             if query[0] in kwargs:
-                raise exc.BadRequest('Attempted to set %s twice.' % query[0])
+                raise exc.BadRequest('Attempted to set %s twice.'
+                                     % query[0].replace('_', '-'))
             kwargs[query[0]] = query[1]
 
         # Make the request to the Ansible Tower API.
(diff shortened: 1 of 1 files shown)


jlaska:
@AlanCoding I like the spirit of the change, however I worry a semi-random `replace` will just move/add confusion elsewhere.  Are we certain that the parameters will *always* use `-`?  Is that enforced by the `click` library.

jlaska:
@AlanCoding I'm luke warm on this change.  What you proposed is a nice low-risk answer.  I'd prefer a tighter coupling between the `tower-cli` parameter names and the tower API.  However, that's beyond the scope of this PR.  Thanks!  Merge if happy.

AlanCoding:
> Is that enforced by the `click` library.  

This is tricky to answer, but my best understanding is that yes, click does this on its own. Give it extra_vars, and the option becomes --extra-vars automatically. (I hope I'm correct on this point)

There are numerous places where tower-cli does a replace on a string for purposes of producing help text with the correct _ vs -. An error message is similar to help text, that's my thinking here. Ideally we would like to get rid of all of those calls to `replace` except for one, but I don't have a concrete proposal to do that right now.

PR #87, opened 2015-10-20
Prepare for release, and example script simplification

AlanCoding:
Some numbers have been updated for the purposes of releasing tower-cli version 2.3.0, such as updating the version number and changing the release date.

In addition, the HISTORY file has been rearranged to be in reverse chronological order. Some other changes were made to the README after testing the format on the testing PyPI server.

Some behaviors were also removed from the example script, so that the example jobs can be run without the need to input any machine credentials.


jlaska:
@AlanCoding looks good, please merge!

Issue #88, opened 2015-10-26
Force on exists gone when creating a project?


mikeycmccarthy:
Hi,

It seems between 2.3.0 and 2.1.1 that force-on-exists has disappeared as an option when creating a project. Is this deliberate?

Thanks
Michael

AlanCoding:
> Is this deliberate?

No. This is a straight-forward bug brought about by other development decisions. Thanks for bringing it to our attention. We can get that straightened out with the next release.

PR #90, opened 2015-11-20
Fix force-on-exists and fail-if-found errors

AlanCoding:
This fixes issue #88, which noted that the --force-on-exists option went missing for creation of a project. That is fixed by adding another class to the list of what `project` inherits from. The bug was caused by my removal of the `create` option from the `MonitorableResource` class (in order to omit it from jobs or ad hoc commands). However, this also removed it from `project` (also monitorable), which led to the subtle error of dropping options defined in the parent class of `create`.

`job_template` had the reverse error due to the same nature of misunderstanding. The options were added to the `create` method in `job_template`, even though it inherited from the standard `Resource` class. This caused the help text to display 2 versions for force-on-exists and fail-on-found. Removing the redundant options was the obvious fix, and that is included in this Pull Request.

Some other changes were made that go along this same vein of consistency with the parent classes. Particularly, the argument lists were edited to follow the same standard as the parent class. These were tested and work fine. Only one method was not an error associated with the 2.3.0 release, and that was the project status method, which was non-functioning if invoked from the command line. Conforming its arguments to the other versions of `status` was done to make it usable, and there are no conflicts with its use in `monitor` to my knowledge.

### DIFF
--- a/lib/tower_cli/resources/job_template.py
+++ b/lib/tower_cli/resources/job_template.py
@@ -17,7 +17,7 @@ from __future__ import absolute_import, unicode_literals
 
 import click
 
-from tower_cli import models, resources
+from tower_cli import models
 from tower_cli.utils import types
 from tower_cli.utils import parser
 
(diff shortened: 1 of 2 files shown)


jlaska:
@AlanCoding :+1:

PR #93, opened 2015-11-23
Bug fix for 2.3.1 release.

AlanCoding:
This is a followup to issue #88. My apologies for merging before the testing was airtight. Both `--force-on-exists` and `--fail-on-found` were still broken for all uses of "create" due to a careless error.

There was a more complicated issue where `--force-on-exists` switches to PATCH instead of POST when an existing project is found. This results in an attempt to PATCH `/organizations/{org pk}/projects/{project pk}/`, which is an invalid endpoint. To fix this, the process for creation of a project belonging to an organization was changed so that a POST is made to `/projects/` and then the `_assoc` method is called right after that. This makes it consistent with the rest of the code base, and provides a resolution to this bug.

With this pull request I am updating the version number to prepare for release because these are fairly important fixes. Let me know if there is anything else I should do differently for the release process.

### DIFF
--- a/lib/tower_cli/models/base.py
+++ b/lib/tower_cli/models/base.py
@@ -994,7 +994,8 @@ class Resource(ResourceMethods):
         do not fail (unless `fail_on_found` is set).
         """
         return super(Resource, self).create(
-            fail_on_found=False, force_on_exists=False, **kwargs
+            fail_on_found=fail_on_found, force_on_exists=force_on_exists,
+            **kwargs
         )
 
     @resources.command
--- a/lib/tower_cli/resources/project.py
+++ b/lib/tower_cli/resources/project.py
@@ -68,26 +68,27 @@ class Resource(models.Resource, models.MonitorableResource):
         This would be a shared class with user, but it needs the ability
         to monitor if the flag is set.
         """
-        backup_endpoint = self.endpoint
+        # First, run the create method, ignoring the organization given
+        answer = super(Resource, self).create(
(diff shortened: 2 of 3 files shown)


jlaska:
Not sure I have a lot to offer on the review front ... :+1:

lib/tower_cli/models/base.py
```diff
@@ -994,7 +994,8 @@ def create(self, fail_on_found=False, force_on_exists=False, **kwargs):
         do not fail (unless `fail_on_found` is set).
         """
         return super(Resource, self).create(
-            fail_on_found=False, force_on_exists=False, **kwargs
```
AlanCoding: This was an artifact from changing the argument list so-many times. It obviously made the flags stop working.

lib/tower_cli/resources/project.py
```diff
@@ -68,26 +68,27 @@ def create(self, organization=None, monitor=False, timeout=None,
...
 
-            self.endpoint = '/organizations/%s%s' % (org_pk, backup_endpoint)
-        answer = super(Resource, self).create(
-            fail_on_found=fail_on_found, force_on_exists=force_on_exists,
-            **kwargs
-        )
-        self.endpoint = backup_endpoint
+            debug.log("associating the project with its organization",
+                      header='details', nl=1)
+            org_resource._assoc('projects', org_pk, project_id)
```
AlanCoding: This uses the _assoc method - equivalent to running `tower-cli organization associate_project ...` command.

PR #94, opened 2015-11-23
Update VERSION to 2.3.1

AlanCoding:
Bug fix version related to problems with the `--force-on-exists` and `--fail-on-found` flags.


AlanCoding:
Updated PR to include a blurb in the release history. Will tag after this is merged.

PR #95, opened 2015-11-24
Inventory Script subcommand added

### LINKED ISSUE #89: Add inventory sub-command to return inventory JSON content
Add @wenottingham for additional comments

AlanCoding:
This adds the `script` subcommand to the inventory resource, it is used in the following way:

> $ tower-cli inventory script 12 --hostvars=1

Unlike other commands, the default format is json, which is done because most foreseeable uses will use this type of output, and the "human" output does not look very polished.

Packaged along with this, we also have tests for 100% coverage of inventory.py, along with some documentation.

I would have preferred to have hostvars default set 1, but there was no simple way to accomplish this while still preserving the user's ability to turn it off (no way which also didn't involve more tests). Thus, I'm going with this as the most elegant solution that we can pull off. If there are any more options to be added, I would love to add them, but the OPTIONS didn't give anything useful and no testing seemed fruitful. The tower-cli command still has many other options which are either appended as baggage in lower-level wrappers, or for the ability to look up an inventory by name.

Fixes #89

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -1,6 +1,10 @@
 Release History
 ===============
 
+2.4.0 (2015-12-xx)
+------------------
+- Added inventory script sub-command
+
 2.3.1 (2015-12-10)
 ------------------
 
@@ -8,6 +12,7 @@ Release History
 - Changed extra_vars behavior to be more compliant by re-parsing vars,
   even when only one source exists
 - Fixed group modify bug, avoid sending unwanted fields in modify requests
+- Added ability to declare user organization admin
 
 2.3.0 (2015-10-20)
 ------------------
--- a/README.rst
+++ b/README.rst
@@ -32,6 +32,8 @@ A few potential uses include:
    Bamboo, etc)
 -  Checking on job statuses
 -  Rapidly creating objects like organizations, users, teams, and more
+-  Pull in `dynamic inventory scripts </docs/inventory_script.rst>`__
+   from Ansible Tower .
 
 Installation
 ------------
--- a/docs/examples/fake_data_creator.sh
+++ b/docs/examples/fake_data_creator.sh
@@ -106,17 +106,17 @@ tower-cli credential create --name=user2 --username=user2 --password=pass1 --tea
 
 echo "Tower-CLI DATA FAKER: creating inventories and groups"
 # Basic localhost examples
-tower-cli inventory create --name=localhost --description="local machine" --organization=Default --variables="variables.yml"
-tower-cli host create --name="127.0.0.1" --description="the host in localhost" --inventory="localhost" --variables="variables.yml"
+tower-cli inventory create --name=localhost --description="local machine" --organization=Default --variables="@variables.yml"
+tower-cli host create --name="127.0.0.1" --description="the host in localhost" --inventory="localhost" --variables="@variables.yml"
 # Corporate example uses localhost with special vars for testing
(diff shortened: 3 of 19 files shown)


wenottingham:
Apologies, this is likely on me for having an incomplete spec.

Inventory scripts are invoked by ansible with very specific arguments/syntax. See http://docs.ansible.com/ansible/developing_inventory.html.

If someone were to want to use this from Ansible, it would seem that they'd need to wrap the tower-cli command in an *additional* script that takes the normal inventory script parameters (--list, --host) and then calls tower-cli with its own appropriate commandline arguments. So we'd either need to add that to Ansible as part of this feature in general to make it turnkey, or we would need to note how you're expected to wrap the above command for use from Ansible.

wenottingham:
When testing this, I get:

```
Unexpected Exception: can't add group to itself
the full traceback was:

  File "/usr/bin/ansible", line 79, in <module>
    sys.exit(cli.run())
  File "/usr/lib/python2.7/site-packages/ansible/cli/adhoc.py", line 111, in run
    inventory = Inventory(loader=loader, variable_manager=variable_manager, host_list=self.options.inventory)
  File "/usr/lib/python2.7/site-packages/ansible/inventory/__init__.py", line 79, in __init__
    self.parse_inventory(host_list)
  File "/usr/lib/python2.7/site-packages/ansible/inventory/__init__.py", line 112, in parse_inventory
    self.parser = get_file_parser(host_list, self.groups, self._loader)
  File "/usr/lib/python2.7/site-packages/ansible/inventory/dir.py", line 58, in get_file_parser
    parser = InventoryScript(loader=loader, groups=groups, filename=hostsfile)
  File "/usr/lib/python2.7/site-packages/ansible/inventory/script.py", line 63, in __init__
    self._parse(stderr)
  File "/usr/lib/python2.7/site-packages/ansible/inventory/script.py", line 135, in _parse
    self.groups[group_name].add_child_group(self.groups[child_name])
  File "/usr/lib/python2.7/site-packages/ansible/inventory/group.py", line 83, in add_child_group
    raise Exception("can't add group to itself")
Exception: can't add group to itself
```

Inventory dump from tower-cli running by hand is at: https://gist.github.com/wenottingham/454414bac19cdc908af6

jlaska:
Are you getting that traceback with tower 2.4.1?

wenottingham:
Yes, happens with both 2.4.0ish and after upgrading to 2.4.1.

AlanCoding:
@wenottingham what do you think is the right way to go about dealing with that error? We could do some work to check for recursive inventory sources in tower-cli and warn the user if there are any, if that is the issue.

Also, would the ultimate goal be to add this as a [module in Ansible core](https://github.com/ansible/ansible/tree/devel/contrib/inventory)? Is the shell script wrapper an acceptable solution? There are other ways we could go about doing that.

I'm wondering if it would be okay to leave this feature for tower-cli version 2.3.2, and leave 2.3.1 for strictly bug fixes.

wenottingham:
I think we should handle those sources somehow - if that inventory works as-is in Tower, it should be output in a manner that it works with Ansible with this script.

In terms of a wrapper to add to Ansible itself in the contrib/inventory directory, would prefer python, but yes, that sounds good.

AlanCoding:
Rebased these changes for the release_2.4 in order to combine with other changes. Will make a new PR that targets the feature branch.

AlanCoding:
It may be better to switch this to use the inventory field type, and cut out as much as possible of the other code. If it is then released, it could be useful later.

PR #97, opened 2015-11-26
Change extra_vars default behavior for 2.3.1 release

### LINKED ISSUE #96: extra vars are not passed to the api properly
in 2.3 if we launch a job with extra args, the extra args are not visible in the job. I think it is because the json is malformed when passed to the api and hence the api ignores the extra args and takes the default in the job template.

AlanCoding:
If this is confirmed to help with the recent issue #96 about extra_vars parsing, then we should include it in the 2.3.1 release. We will merge this PR, tag, and then release if so.

Fixes #96

### DIFF
--- a/lib/tower_cli/resources/job_template.py
+++ b/lib/tower_cli/resources/job_template.py
@@ -82,10 +82,10 @@ class Resource(models.Resource):
         You may only include one --extra-vars flag with this command, and
         whatever you provde will overwrite the existing field. Start this
         with @ in order to indicate a filename."""
-        if 'extra_vars' in kwargs:
+        if kwargs.get('extra_vars'):
             # read from file, if given
             kwargs['extra_vars'] = \
-                parser.file_or_yaml_split(kwargs['extra_vars'])
+                parser.extra_vars_loader_wrapper([kwargs['extra_vars']])
         return super(Resource, self).modify(
             pk=pk, create_on_missing=create_on_missing, **kwargs
         )
--- a/lib/tower_cli/resources/job_template.py
+++ b/lib/tower_cli/resources/job_template.py
@@ -58,11 +58,14 @@ class Resource(models.Resource):
     job_tags = models.Field(required=False, display=False)
     skip_tags = models.Field(required=False, display=False)
     extra_vars = models.Field(required=False, display=False)
+    ask_variables_on_launch = models.Field(
+        type=bool, required=False, display=False,
+        help_text='Prompt user for extra_vars on launch.')
     become_enabled = models.Field(type=bool, required=False, display=False)
 
     @click.option('--extra-vars', required=False, multiple=True,
-                  help='yaml format text that contains extra variables '
-                       'to pass on. Use @ to get these from a file.')
+                  help='Extra variables used by Ansible in YAML or key=value '
+                       'format. Use @ to get YAML from a file.')
     def create(self, fail_on_found=False, force_on_exists=False,
                extra_vars=None, **kwargs):
         """Create a job template.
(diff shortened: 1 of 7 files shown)


AlanCoding:
There's a lot of material here, and a lot of tests to go along with them. I will try to give some summarization and background. Firstly, the first 2 commits here changed enough code to solve the problem raised in the issue, and that's all. The 3rd commit was performing a broader audit of consistency with the UI behavior. tower-cli extra_vars were already designed for consistency with the behavior of Ansible core, so there was a balancing act of being more like one versus the other.

The UI behavior:

- **Job Template creation/editing** - parses JSON or YAML in-browser according to the user selection. Confirming that the input is valid is done inside of this view, and the text is passed as-is when saved.
- **Job launches without special considerations** - In the current version, `extra_vars` are absent from the payload
- **Job launches with prompt for extra_vars** - The JT `extra_vars` are presented to the user, and the user is free to edit them, but it is converted to JSON when the POST request is made

So what does this mean for tower-cli? I don't think I'm going out on a limb when I say that tower-cli should throw an error if the user gives it invalid data. However, being like Ansible, it should still eat up multiple `--extra-vars` options, which can include files too (which also might have YAML and comments). You can go very deep into the corner cases through this line of logic.

The new tower-cli behavior:

- **Job Template creation/modification** - Will take all the extra-vars given and try to append them in YAML form. It that can't be done, it gives up and submits it as JSON. This is somewhat similar to how the UI behaves.
- **Job launches without special considerations** - Combines the multiple extra-vars flags, but dumps them into JSON.
- **Job launches with prompt for extra_vars** - Converts the JT extra_vars into YAML (without curly braces) and puts that into the editor for the user to modify. If the user edits this into valid YAML with comments, then we send that. It will still error if something goes wrong.
- **Flags versus files** - the key=value syntax will not be allowed for content from files.

tests/test_resources_job.py
```diff
@@ -131,15 +132,35 @@ def test_extra_vars_at_runtime(self):
...
+                'id': 1,
+                'name': 'frobnicate',
+                'related': {'launch': '/job_templates/1/launch/'},
+            })
+            with mock.patch.object(click, 'edit') as edit:
+                edit.return_value = '# Nothing.\nfoo: bar'
+                result = self.res.launch(
+                    1, no_input=False
+                )
+                self.assertIn('# comment', edit.mock_calls[0][1][0])
```
AlanCoding: This tests to make sure that if the Job Template extra_vars contain comments, those are displayed to the user when prompting for extra_vars. The Job Template extra_vars could be JSON, so that's a different test case.

lib/tower_cli/resources/job_template.py
```diff
@@ -58,11 +58,14 @@ class Resource(models.Resource):
     job_tags = models.Field(required=False, display=False)
     skip_tags = models.Field(required=False, display=False)
     extra_vars = models.Field(required=False, display=False)
+    ask_variables_on_launch = models.Field(
+        type=bool, required=False, display=False,
+        help_text='Prompt user for extra_vars on launch.')
```
AlanCoding: Note: I did add another command line option for Job Template here. That's because several fixes were applied to job launch when extra_vars prompt was enabled, and you need this in order to enable the prompt through tower-cli. You can go back and fourth between interfaces, but I would rather not leave it that way since it's a low-risk addition.

lib/tower_cli/utils/parser.py
```diff
@@ -83,40 +86,28 @@ def string_to_dict(var_string):
...
-            assert type(return_dict) is dict
-        except (AttributeError, yaml.YAMLError, AssertionError):
-            # if these fail, parse by key=value syntax
-            try:
-                return_dict = parse_kv(var_string)
-            except:
-                raise exc.TowerCLIError(
-                    'failed to parse some of the extra '
-                    'variables.\nvariables: \n%s' % var_string
-                )
```
AlanCoding: Big change: the JSON parser was removed here. That was done to avoid unicode issues. You can't expect to load data with the JSON parser and then dump it from the YAML parser (which you would necessarily do, given the specs of this change), so the only option is to load both types with the YAML parser, and this seems to work fine for all forms of testing with no ill effects.

PR #99, opened 2015-12-03
Modify command bugs resolved

### LINKED ISSUE #98: Bugs in modify command
The following command fails in a way which is obviously in error. Additionally, it makes a lot of unexpected requests related to the inventory source that should not happen.

> tower-cli group modify --name=EC2 --description "has been modified"  -v

Another issue is that the following command does not work as it should.

> tower-cli job_template modify --name="Apache" --description "has been modified" -v

It sends `{'description': u'has been modified', 'job_type': u'run', 'name': u'Apache'}` in the PATCH request. The `job_type` field is an artifact of the way that tower-cli processes default values. In this case, it would override any other selection for the field in the process of modifying something unrelated. This is a silent bug that could disrupt other processes that the command is used in.

AlanCoding:
As referenced in the respective issue, this deals mainly with two problems. 1) `tower-cli group modify` will error 2) `tower-cli job_template modify` will silently modify things that it shouldn't. These both relate to hazards with setting defaults that also apply to the `modify` command. To explain that better, I would distinguish between command-line-level defaults, and defaults for a specific method. When you set a default on a field inside of a class, it's not exactly clear what the intention is.

# Group Modify Fix

This command takes two actions - first it modifies the group, then it modifies the group's connected inventory_source. The problem is that the `--source` option on the group modify commands was set to an invalid value as a default, so so it would fail without that option specified. Since this was implemented with a click.option decorator, it can just be removed from the modify command and not the create command.

Another minor issue is that the result would appear green (represents unchanged) when it should have appeared yellow (changed). That is because it modifies two things in the method. A a quick fix, I set it to output whenever of the two changed. If they both changed, it doesn't matter much, because the information displayed to the user isn't very different between the two.

# Job Template Fix

Previously, `job_type` had its default set to `run`. You can still create a JT and set job_type to check, but if you then edit something unrelated in this JT (like its description), it would set the job_type back to run. We would like to change the logic more broadly so that we can still use field defaults in this way, but it doesn't go well. If a JT has its job_type set to check, and then it is modified, there is no way for the method `write` in `base.py` to tell if that was a user input, or if it was passed as a default. Thus, we can't really fix the issue on this level.

The simple solution is to have a special clause in creation of a job template that sets the job_type if none is given. That is the solution I went with.

Fixes #98

### DIFF
--- a/lib/tower_cli/resources/group.py
+++ b/lib/tower_cli/resources/group.py
@@ -58,18 +58,19 @@ class Resource(models.Resource):
         """
         # Break out the options for the group vs its inventory_source
         group_fields = [f.name for f in self.fields]
-        inv_kwargs = {}
+        is_kwargs = {}
         for field in kwargs.copy():
             if field not in group_fields:
-                inv_kwargs[field] = kwargs.pop(field)
+                is_kwargs[field] = kwargs.pop(field)
 
         # Handle alias for "manual" source
-        if 'source' in inv_kwargs and inv_kwargs['source'] == 'manual':
-            inv_kwargs.pop('source')
+        if 'source' in is_kwargs and is_kwargs['source'] == 'manual':
+            is_kwargs.pop('source')
 
         # First, create the group.
         answer = super(Resource, self).create(
-            fail_on_found=False, force_on_exists=False, **kwargs)
+            fail_on_found=fail_on_found, force_on_exists=force_on_exists,
+            **kwargs)
 
         # If the group already exists and we aren't supposed to make changes,
         # then we're done.
@@ -80,7 +81,7 @@ class Resource(models.Resource):
         # with the inventory source at all? If no credential or source
         # was specified, then we'd just be updating the inventory source
         # with an effective no-op.
-        if len(inv_kwargs) == 0:
+        if len(is_kwargs) == 0:
             return answer
 
         # Get the inventory source ID ("isid").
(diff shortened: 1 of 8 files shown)


AlanCoding:
The last commit was solving another issue related to groups and inventory_source. The tower API expects an empty string, "", for when the source is set to manual. So I just added one "if" statement that changed "manual" to "" inside of `group.py`. The inventory_source resource just has its manual option set to an empty string (because it's unlikely that a user will ever interact directly with it anyway).

It fixes the problem, with this, you can now modify a group and set it back to a manual source from a cloud source. In the process of creation, the bug never surfaced because of other special cases.

lib/tower_cli/resources/group.py
```diff
@@ -52,23 +52,36 @@ class Resource(models.Resource):
...
         """
+        # Break out the options for the group vs its inventory_source
+        group_fields = [f.name for f in self.fields]
+        is_kwargs = {}
+        for field in kwargs.copy():
+            if field not in group_fields:
+                is_kwargs[field] = kwargs.pop(field)
+
+        # Handle alias for "manual" source
+        if 'source' in is_kwargs and is_kwargs['source'] == 'manual':
```
chrismeyersfsu: You could make this more "compact" with  `if is_kwargs.get('source', None) == 'manual'`

lib/tower_cli/resources/group.py
```diff
@@ -52,23 +52,36 @@ class Resource(models.Resource):
...
+            is_kwargs.pop('source')
```
chrismeyersfsu: or you could do `is_kwargs.pop('source', None') if is_kwargs.get('source', None) == 'manual' else None` <-- but that is more confusing

lib/tower_cli/resources/group.py
AlanCoding: Right, this syntax follows the tower-cli standard better. I'll definitely change these instances.

chrismeyersfsu: It's a preference thing. No big deal.

chrismeyersfsu:
:+1:

PR #105, opened 2015-12-11
Standard out functionality

### LINKED ISSUE #74: Include `result_stdout` in job get output
Users should be able to view the `result_stdout` from a job (potentially includes project_update, inventory_update, system_job, ad_hoc_command, and job).

The `result_stdout` should be included in the output from the `get` sub-command.  `tower-cli` should handle cases where the `result_stdout` is too large, and use the "Download" API feature.

AlanCoding:
Streaming standard out:

    tower-cli job launch --job-template=74 --monitor --stdout

Getting the standard out from a job that has already ran:

    tower-cli job stdout 572

This functionality is focused very heavily on a human user running tower-cli from the command line. It also gets us somewhat closer to Ansible core parity, because it allows you to kick off a job and see the standard out stream back to you as it goes. Of course, it's a little more complicated with Tower because it will sit in the queue before it can start, and other factors like that. The testing of this so-far has consisted of trial and error getting the output to format exactly like what we want. However, if the `--stdout` flag is not given, there should not be any possibility of negatively affecting other features of tower-cli.

The unit test situation is going to be a little complicated, so I haven't even attempted to write tests just yet. Since all of the streaming exists inside of a loop in the monitor method, it is very difficult to try all of the permutations in the job unit tests without setting up a vast universe of mock. We will need some feedback on that, so I'm marking this "in progress".

Fixes #74

### DIFF
--- a/lib/tower_cli/models/base.py
+++ b/lib/tower_cli/models/base.py
@@ -25,6 +25,7 @@ import sys
 import time
 from copy import copy
 from sdict import adict
+from base64 import b64decode
 
 import six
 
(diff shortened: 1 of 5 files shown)


AlanCoding:
Moving to release_2.4

PR #107, opened 2015-12-12
Permission feature added

### LINKED ISSUE #51: Creating permissions through Tower CLI
We'd love to be able to create permissions through Tower CLI. Our playbooks to bootstrap Tower are a mix of CLI calls and API calls. Creating permissions (and other objects) is fine through the API (really liking the browseable documentation) but there are far more steps to do something idempotent in true Ansible style (lookup the permission, POST if not there, PUT if there etc vs just a tower-cli --force-if-exists). 

Thanks!

AlanCoding:
This adds the ability to create and manage permissions through tower-cli. This has some unique challenges with it - because you can not list permissions globally by a request to `/permissions/`. The ability to do was baked in as an assumption to many common tasks, so this had to be teased out as a special case for the permission class. Instead of doing things globally, you have to specify either a user or a team for every action you take. Because of that, you'll see this code do a switcharoo of the endpoints  before/after the permission primary key is established. That is because of the added complication that you can't subsequently make any requests to and endpoint like `/users/10/permissions/5/`. That is what several tower-cli methods would automatically construct, so special care is given to that case to turn it into `/permissions/5/`.

I haven't yet gone down the full road of testing, because I thought a fast feature sprint might be more efficient in the long run, because this could be tested along side the other new features. The commands that I have used and have found to generally work with consistency with other tower-cli behavior are:

```bash
tower-cli permission create --name="from_CLI" --user=18 --inventory=19 -v
tower-cli permission modify --name="from_CLI" --user=18 --description="edit from cli" -v
tower-cli permission delete --name="from_CLI" --user=bio_citizen -v
tower-cli permission list --user=bio_citizen
tower-cli permission get --name=righttoclick --user=bio_citizen
```

Fixes #51

### DIFF
--- a/lib/tower_cli/resources/permission.py
+++ b/lib/tower_cli/resources/permission.py
@@ -23,6 +23,7 @@ class Resource(models.Resource):
     cli_help = 'Manage permissions within Ansible Tower.'
     endpoint = '/permissions/'
     identity = ('name', )
+    no_lookup_flag = False
 
     # Permissions must be created for either a user or a team
     name = models.Field(unique=True, required=False, display=True)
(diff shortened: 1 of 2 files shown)


AlanCoding:
Remains to be done:

- More detailed command testing
- Unit tests

If nothing else is found to be broken, that should be all.

AlanCoding:
Moving to release_2.4 feature branch

AlanCoding:
Even though this applies to pre-3.0 releases, I'm going to put in in for the tower-cli 3.0 release.

PR #108, opened 2015-12-12
Accept file or direct variables for inventory, host, group

### LINKED ISSUE #104: Use "@" syntax to denote file name for all variable fields
`Job` launch and `Job Template` actions can accept a file name by starting the string with a `@` character, otherwise, it will take the variables given directly in the command line. Consider the following examples, where both commands pass the extra variable a=5 to the job launch, and a_equals_5.json is a file that contains the same text.

    tower-cli job launch --job-template=5 --extra-vars='@a_equals_5.json'
    tower-cli job launch --job-template=5 --extra-vars='{"a": 5}'

However, some other flags (for variables) can only specify a file. In other words, to set a=5, you have to create a file (such as a_equals_5.json) with that content and use the flag `--variables=@a_equals_5.json'`. These include:

- `--variables` flag on inventory
- `--variables` flag on group
- `--variables` flag on host
- `--source-vars` flag on group and inventory_source

There are several reasons the updated syntax is more usable. Saving a file often constitutes extra legwork and leaves junk lying around. It's also a quick route to passing local environment variables to something that resides in Tower. This would also be easy to implement in the next release.

AlanCoding:
This will allow the `--variables` option on inventory, host, group (and `--source-vars` on group) to retrieve data from a file if it starts with an "@". This PR is mainly doing routing work through the click library capabilities combined with the specific tower-cli architecture. Variables are implemented as a custom type. I wanted to allow multiple flags (like in job and job template), but I couldn't quite work through the technical challenges of doing that. This is still a much more elegant approach than what I have been using, and it shows a minimal footprint on the code for the resources (which is a part of the tower-cli design philosophy).

I need to check back later and audit what's showing for help text, as well as do more testing, so putting the in_progress label on this for now.

Fixes #104

### DIFF
--- a/docs/examples/fake_data_creator.sh
+++ b/docs/examples/fake_data_creator.sh
@@ -106,17 +106,17 @@ tower-cli credential create --name=user2 --username=user2 --password=pass1 --tea
 
 echo "Tower-CLI DATA FAKER: creating inventories and groups"
 # Basic localhost examples
-tower-cli inventory create --name=localhost --description="local machine" --organization=Default --variables="variables.yml"
-tower-cli host create --name="127.0.0.1" --description="the host in localhost" --inventory="localhost" --variables="variables.yml"
+tower-cli inventory create --name=localhost --description="local machine" --organization=Default --variables="@variables.yml"
+tower-cli host create --name="127.0.0.1" --description="the host in localhost" --inventory="localhost" --variables="@variables.yml"
 # Corporate example uses localhost with special vars for testing
-tower-cli inventory create --name=Production --description="Production Machines" --organization="Hyrule Ventures" --variables="variables.yml"
+tower-cli inventory create --name=Production --description="Production Machines" --organization="Hyrule Ventures" --variables="@variables.yml"
 tower-cli group create --name=EC2 --credential="AWS creds" --source=ec2 --description="EC2 hosts" --inventory=Production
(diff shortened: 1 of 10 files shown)


AlanCoding:
For these half-complete features, I wanted to document what remains.

Remains to be done:

- Should have made `--source-vars` on group the `Variable` type as well.
- Allow multiple flags
- More detailed command testing
- Unit tests

The multiple flags feature is the real tricky one here. Click combines multiple flags into a tuple. I looked into it, and it would appear that all we need is to override `type_cast_value` on the Option class.

https://github.com/mitsuhiko/click/blob/6fcda3e5fcddd9001b314e83a1e0546bd22c78c2/click/core.py#L1319

Now, part of the trickery is that our custom types override `click.types.ParamType`, not `click.option`. Even click.option actually comes from a wrapper that lives at `click.decorators.option`. Mucking with click.option is still perfectly fine with me, and this is logical since the combination into a tuple occurs at a lower level than the parameter type definition. But this is still more ambitious than the norm.

Additionally, the ability to provide a custom `Option` class appears to be a specific [feature](https://github.com/mitsuhiko/click/commit/8e2e583694ac0b374a3aa9712e8274ef3519f1d1) of click. So the door is open to overwrite what I'm thinking about, I just have to make sure this is the best practice.

AlanCoding:
Switching this pull request to the release_2.4 feature branch.

PR #110, opened 2015-12-16
Apply variable processing to group, host, and inventory options

### LINKED ISSUE #104: Use "@" syntax to denote file name for all variable fields

AlanCoding:
This replaces Pull #108, and switches it to the feature branch. I believe that linking to the issue will not merge it when this is incorporated into the feature branch, but will close when the feature branch is merged into master. The description is copied from there:

--
This will allow the `--variables` option on inventory, host, group (and `--source-vars` on group) to retrieve data from a file if it starts with an "@". This PR is mainly doing routing work through the click library capabilities combined with the specific tower-cli architecture. Variables are implemented as a custom type. I wanted to allow multiple flags (like in job and job template), but I couldn't quite work through the technical challenges of doing that. This is still a much more elegant approach than what I have been using, and it shows a minimal footprint on the code for the resources (which is a part of the tower-cli design philosophy).

I need to check back later and audit what's showing for help text, as well as do more testing, so putting the in_progress label on this for now.

Fixes #104

### DIFF
--- a/docs/examples/fake_data_creator.sh
+++ b/docs/examples/fake_data_creator.sh
@@ -106,17 +106,17 @@ tower-cli credential create --name=user2 --username=user2 --password=pass1 --tea
 
 echo "Tower-CLI DATA FAKER: creating inventories and groups"
 # Basic localhost examples
-tower-cli inventory create --name=localhost --description="local machine" --organization=Default --variables="variables.yml"
-tower-cli host create --name="127.0.0.1" --description="the host in localhost" --inventory="localhost" --variables="variables.yml"
+tower-cli inventory create --name=localhost --description="local machine" --organization=Default --variables="@variables.yml"
+tower-cli host create --name="127.0.0.1" --description="the host in localhost" --inventory="localhost" --variables="@variables.yml"
 # Corporate example uses localhost with special vars for testing
-tower-cli inventory create --name=Production --description="Production Machines" --organization="Hyrule Ventures" --variables="variables.yml"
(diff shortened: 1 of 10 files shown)


AlanCoding:
This might have to get rolled back. It looks like there might be valid cases where the source_vars might be something like `tag:Name=webservers1`. If that's the case, then this change would prevent some users from being able to use the feature at all.

The custom click.option should probably still stick around. But it needs to be rolled out more cautiously. For the code changes here, we might apply it to 1 or 2 out of the 3 cases of variable-list fields.

PR #118, opened 2015-12-18
Use the variables field type to consolidate code

AlanCoding:
There was a lot of extra legwork being done in the job template that we can now remove with use of the variable type field. This PR removes all those extra methods on job template that we can really do without.


AlanCoding:
This doesn't make as much sense without the other branches before it, so I'm going to close it as I or-orient all the features to master.

PR #127, opened 2016-02-24
Small flake8 fix to make the tests pass again

AlanCoding:
This is a non-functional movement of one line. I hate to make an entire PR out of it, but we need this merged if we want any of the other pull requests to show that they're passing again.

But first, I want to make sure this gets all checks.


anoek:
:+1:

PR #132, opened 2016-03-18
More robust pagination logic

AlanCoding:
This shouldn't affect most users, but if Tower gives a different format for the first page, this will allow it to work correctly. Corresponding test documents the particular case.


jlaska:
@cchurch any thoughts with how the page#1 link differs with more recent versions of Django?

wwitzel3:
Can't speak to the specific Django/DRF issue, but the code and test LGTM :+1:

cchurch:
This change was introduced by updates to Django REST Framework.  Since `/api/v1/whatevers/` and `/api/v1/whatevers/?page=1` are equivalent, I don't see any harm in supporting both.

PR #135, opened 2016-04-01
Add ability to create group as subgroup of another group

AlanCoding:
This allows the following type of command to work where `parent_group` is an existing group and the `child_group` is created.

    tower-cli group create --parent-group=parent_group --name=child_group

This also introduces an error scenario where neither inventory nor group is provided (previously we just returned the error from the server when inventory was not there).

### DIFF
--- a/lib/tower_cli/resources/group.py
+++ b/lib/tower_cli/resources/group.py
@@ -226,6 +226,20 @@ class Resource(models.Resource):
         isid = self._get_inventory_source_id(group)
         return isrc.update(isid, monitor=monitor, timeout=timeout, **kwargs)
 
+    @resources.command(use_fields_as_options=False)
+    @click.option('--group', type=types.Related('group'))
+    @click.option('--parent', type=types.Related('group'))
+    def associate(self, group, parent):
+        """Associate this group with the specified group."""
+        return self._assoc('children', parent, group)
+
+    @resources.command(use_fields_as_options=False)
+    @click.option('--group', type=types.Related('group'))
+    @click.option('--parent', type=types.Related('group'))
+    def disassociate(self, group, parent):
+        """Disassociate this group from the specified group."""
+        return self._disassoc('children', parent, group)
+
     def _get_inventory_source_id(self, group):
         """Return the inventory source ID given a group dictionary returned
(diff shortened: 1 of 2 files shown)


AlanCoding:
This has a flaw in its implementation - group names are not globally unique. The parent group may need to also have the inventory specified. If a group with that name exists in multiple groups, then it would be impossible to add children to it (in the current state of this PR).

Instead, we're going to have to:

 - Ditch the use of related type
 - Add a custom lookup function that takes group and inventory
 - Selectively check if user provided inventory, and make appropriate decisions

The new lookup function will be easily reused for listing the children in a group (which is a standing issue).

AlanCoding:
At this point, the basic mechanics for incorporation of inventory into the group management commands is working, and the other PR is merged in, and also integrated to the use of inventory. I'm going to share my testing bash script here:

```
# Setup information
tower-cli organization create --name Default
tower-cli inventory create --name inv1 --organization Default
tower-cli inventory create --name inv2 --organization Default

# Creating simple parent and child structure
tower-cli group create --name group_top --inventory inv1
tower-cli group create --name group_child --parent group_top
# Check listin of top group's children and see the child
tower-cli group list --parent group_top

# Make sure same commands work with a group of the same name
#  when that group is inside of a second inventory
tower-cli group create --name group_top --inventory inv2
tower-cli group create --name group_child --parent group_top --inventory inv1
tower-cli group list --parent group_top --inventory inv1

# Drop a group down into top level
tower-cli group create --name top_bottom --parent group_top --inventory inv1
tower-cli group disassociate --parent group_top --group top_bottom --inventory inv1
# this list should show the "top_bottom" group because of --root option
tower-cli group list --inventory inv1 --root

# Bring that group back into the parent group as a child
tower-cli group associate --parent group_top --group top_bottom --inventory inv1
# This should show that "top-bottom" has moved back into the group_top
tower-cli group list --parent group_top --inventory inv1
```

This script articulates all of the cases that I initially had in mind for this to work for. I might still think of more things later.

We could also use some better help text documentation (although the status of fields makes this tricky). Also, I haven't scratched the surface for unit tests with the inventory flag yet.

Issue #136, opened 2016-04-15
Passing tags to tower-cli results in started job without tags


rvdh:
Here I launch a new job based on template 272 with the tags default,npm,pm2,env:
```
$ ~/webhook/bin/tower-cli job launch --job-template 272 --tags default,npm,pm2,env -v -f json
*** DETAILS: Getting the record. **********************************************
GET https://192.168.4.14/api/v1/job_templates/272/
Params: {}

*** DETAILS: Asking for information necessary to start the job. ***************
GET https://192.168.4.14/api/v1/job_templates/272/launch/
Params: {}
```
Here you can see it's launching the job with the job_tags data:
```
*** DETAILS: Launching the job. ***********************************************
POST https://192.168.4.14/api/v1/job_templates/272/launch/
Data: {u'job_tags': u'default,npm,pm2,env'}
```
But when asking for the job status, you'll see the tags have gone missing (see "job_tags"):
```
*** DETAILS: Asking for job status. *******************************************
GET https://192.168.4.14/api/v1/jobs/15270/
Params: {}

{
  "id": 15270,
  "type": "job",
  "url": "/api/v1/jobs/15270/",

< snip non-interesting json >

  "created": "2016-04-15T11:13:16.995Z",
  "modified": "2016-04-15T11:13:17.055Z",
  "name": "dev.exonet.nl deploy",
  "description": "",
  "job_type": "run",
  "inventory": 1,
  "project": 88,
  "playbook": "projects_deployment/main.yml",
  "credential": 1,
  "cloud_credential": null,
  "forks": 0,
  "limit": "",
  "verbosity": 3,
  "extra_vars": "{}",
  "job_tags": "",
  "force_handlers": false,
  "skip_tags": "",
  "start_at_task": "",
  "unified_job_template": 272,
  "launch_type": "manual",
  "status": "pending",
  "failed": false,
  "started": null,
  "finished": null,
  "elapsed": 0.0,
  "job_args": "",
  "job_cwd": "",
  "job_env": {},
  "job_explanation": "",
  "result_stdout": "Waiting for results...",
  "result_traceback": "",
  "job_template": 272,
  "passwords_needed_to_start": [],
  "ask_variables_on_launch": false,
  "changed": true
}
```
If I look at the REST api, there's no such thing as tags for the launch command of a job_template:

```
OPTIONS /api/v1/job_templates/272/launch/
HTTP 200 OK
Content-Type: application/json
Vary: Accept
[...]

PR #137, opened 2016-04-16
Allow starting job with job_tags for superadmins

styxit:
Allow super admins to start jobs through  `POST /jobs` and `GET /jobs/%d/start`. Which allows the use of providing tags. For other users, use the template `/launch` endpoint.

In order to do this, a call to `/me` had to be made to determine if a user has superuser privileges.

If this is not the way to go, feel free to close this pull request.

See issue: ansible/tower-cli#136

### DIFF
--- a/lib/tower_cli/api.py
+++ b/lib/tower_cli/api.py
@@ -47,7 +47,7 @@ class Client(Session):
             adapter.max_retries = 3
 
         # Store the authenticated user details.
-        #self.me = (self.get('/me').json())['results'][0]
+        self.me = (self.get('/me').json())['results'][0]
 
     @property
     def prefix(self):
--- a/lib/tower_cli/api.py
+++ b/lib/tower_cli/api.py
@@ -39,15 +39,10 @@ class Client(Session):
 
       [1]: http://docs.python-requests.org/en/latest/
     """
-    me = {}
-
     def __init__(self):
         super(Client, self).__init__()
         for adapter in self.adapters.values():
             adapter.max_retries = 3
-
-        # Store the authenticated user details.
-        self.me = (self.get('/me').json())['results'][0]
 
     @property
     def prefix(self):
--- a/lib/tower_cli/api.py
+++ b/lib/tower_cli/api.py
@@ -39,10 +39,15 @@ class Client(Session):
 
       [1]: http://docs.python-requests.org/en/latest/
     """
+    me = {}
+
     def __init__(self):
         super(Client, self).__init__()
         for adapter in self.adapters.values():
             adapter.max_retries = 3
+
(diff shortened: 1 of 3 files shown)


AlanCoding:
Thanks for the contribution!

I see what you're doing with the `/me/` endpoint, and that's fine by me as a means of determining if the user is a superuser. However, this would direct launch commands with the `--job-template` flag to the `/jobs/` launch method. In many cases, the job template will have configuration data that the user is counting on.

My initial thoughts here are to:

 - add a clause to always use the job template launch if the `--job-template` flag is given
 - if that flag is not given, then proceed with the superuser check logic in this PR

styxit:
> if that flag is not given, then proceed with the superuser check logic in this PR

In that case, how can a job be created if the only parameters for this command are:

```
  def launch(self, job_template=None, tags=None, monitor=False, timeout=None,
               no_input=True, extra_vars=None, **kwargs):
```

AlanCoding:
Having the `job_template` parameter without any others is sufficient to launch a job. There's not a problem that other parameters like tags or extra_vars are None in that case.

rvdh:
@AlanCoding 

>  However, this would direct launch commands with the --job-template flag to the /jobs/ launch method. 
> In many cases, the job template will have configuration data that the user is counting on.

Indeed, that's why the /jobs/ POST call supports a `job_template` field, to create a new job based on a template, right?

> My initial thoughts here are to:
> 
> add a clause to always use the job template launch if the --job-template flag is given
> if that flag is not given, then proceed with the superuser check logic in this PR

If that flag is not given, you can't create a new job based on a template. So thats not what we're trying to achieve here - we just want to run a job, based on a template, **with** tags.

AlanCoding:
> Indeed, that's why the /jobs/ POST call supports a job_template field, to create a new job based on a template, right?

Thanks for bringing this up, I realize that I was wrong about what the intent was. `/jobs/` provides a great amount of flexibility for superusers to create jobs on the fly with any amount of over-writing of the Job Template fields.

We still want to default to the `/launch/` endpoint, and to enable the old use we need a new scheme to distinguish between the two in a way that is predictable to the user. I will double-check with the docs and come back to this topic.

AlanCoding:
We have implemented another way to get this functionality in PR #184. The concern was that logic to deal with a POST to the `/jobs/` endpoint will affect jobs launched via a job template as well, so in the new PR, the "old" way of launching jobs is enabled by the use of a flag. This will allow the passing of tags in the launch, as well as other fields.

AlanCoding:
We should have this functionality covered, and will be releasing it within the next week or so. Please don't hesitate to ask if you want to know how to do a particular task with tower-cli in that release.

PR #138, opened 2016-04-22
Allow associate/disassociate on group to allow for child groups.

wenottingham:
Usage: tower-cli group associate --group "foo" --parent "bar"

### DIFF
--- a/lib/tower_cli/resources/group.py
+++ b/lib/tower_cli/resources/group.py
@@ -196,6 +196,20 @@ class Resource(models.Resource):
         isid = self._get_inventory_source_id(group)
         return isrc.update(isid, monitor=monitor, timeout=timeout, **kwargs)
 
+    @resources.command(use_fields_as_options=False)
+    @click.option('--group', type=types.Related('group'))
+    @click.option('--parent', type=types.Related('group'))
+    def associate(self, group, parent):
+        """Associate this group with the specified group."""
+        return self._assoc('children', parent, group)
+
+    @resources.command(use_fields_as_options=False)
+    @click.option('--group', type=types.Related('group'))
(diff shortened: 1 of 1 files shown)


AlanCoding:
Do you see this as something complementary to the group creation method in #135, or replacing it? We would still use the child list capability from that either way.

I had found that cross-inventory scenarios can complicate the situation for managing the group hierarchy a lot. I generally like your approach here. It's simple and it will work in most cases. For a corner case, however:

> tower-cli group associate --group "foo" --parent "bar"

Consider the case where the `bar` group exists inside of multiple inventories. I think it will fail relatively nicely (error that it found 2 bars, can't do that). We might want to move toward new custom field types, but I didn't see an implementation that could work for this type of thing. The command should, in theory, take just 1 inventory flag, because if either foo or bar are not in that inventory, the request should be rejected.

wenottingham:
It accomplishes the same thing as #135, I just didn't search first (my bad).  Maybe adding an optional --inventory parameter that is used for differentiation in the case that there are multiple groups of name 'foo' or 'bar' per the answer above.

AlanCoding:
It doesn't solve the exact same problem, and I wanted dis / associate functions for this. Creating the tree structure and managing it are different things. I can merge them together.

> Maybe adding an optional --inventory parameter that is used for differentiation in the case that there are multiple groups of name 'foo' or 'bar' per the answer above.

I like that, and I'm fine doing a one-off solution by copying the approach I used in the `create` method. That is, changing the argument type and adding inventory to the `get` if it is present. I still want to allow the syntax you initially presented, and if the user gets it wrong, then that's in the realm of exception handling.

AlanCoding:
This has now been fully integrated into the other PR for managing group nesting. Closing this PR.

PR #140, opened 2016-05-01
Concept to fix inability to disassociate things

### LINKED ISSUE #133: Can not switch ownership of credential from team to user
Consider that I have a credential (pk=6) that is owned by a team. I want to use tower-cli to switch the ownership to a user. There appears to be no way to do this. If I try to specify "no team", it will not accept that because it is interpreted as "associate the credential with the team named `null`".

```bash
$ tower-cli credential modify 6 --user=1 --team=null
Error: Could not get team. The requested object could not be found.
```

If I don't include the team, then tower-cli tries to modify the user field *in addition to* what already exists. This causes a validation problem.

```bash
$ tower-cli credential modify 6 --user=1

PATCH http://192.168.99.100:8013/api/v1/credentials/6/
Params: None
Data: {"user": 1}

Response: {"__all__":["Credential cannot be assigned to both a user and team"]}
```

There should be some designator that tells tower-cli to deallocate the team or user field, and this should be pointed out in the help text.

AlanCoding:
This would let people move credentials from user to team ownership, and in general allow telling the API that a field should be "null". The syntax to specify this would be:

> tower-cli resource command 14 --related-field=null

Doing this is a valid action. You don't often have the ability to make that change, but that's aside from the point. The linked issue is a rare scenario where it is absolutely needed.

Fixes #133

### DIFF
--- a/lib/tower_cli/models/base.py
+++ b/lib/tower_cli/models/base.py
@@ -581,6 +581,11 @@ class ResourceMethods(BaseResource):
             answer.update(existing_data)
             return answer
 
+        # Reinsert None for special case of null association
+        for key in kwargs:
+            if kwargs[key] == 'null':
+                kwargs[key] = None
+
         # Get the URL and method to use for the write.
         url = self.endpoint
         method = 'POST'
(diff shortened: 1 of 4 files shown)


wwitzel3:
We should probably handle this better on the Tower API side, but this is a nice work around in the meantime. 👍

PR #141, opened 2016-05-01
SSL corner case error and suggestion

### LINKED ISSUE #121: Discussion needed on how to best handle non ssl connections
Currently all connections default to using https. For systems that do not have https setup, a workaround of setting the host to `http://<ip>:<port>/` is possible, but it seems like maybe we should have a config option and/or flag for setting the protocol. Perhaps call it `protocol` with the default value being `https`?

AlanCoding:
This adds some verbosity to tower-cli in order to prevent a subset of problems that users have encountered with SSL errors.

Firstly, we are going to raise an error for the obviously wrong combination of an `http` host with verify_ssl set.

That still leaves the other possibility that tower-cli automatically attached `https` to the host, against the user's intent. This is tricky, and my past attempts to filter out the error did not work most of the time. So instead, we will tell the user when a SSLError comes up and this _could_ be the case. In retrospect, this looks like the best approach.

Fixes #121

### DIFF
--- a/lib/tower_cli/api.py
+++ b/lib/tower_cli/api.py
@@ -25,7 +25,7 @@ from requests.models import Response
 from requests.packages import urllib3
 
 from tower_cli.conf import settings
-from tower_cli.utils import data_structures, debug, exceptions as exc
+from tower_cli.utils import data_structures, debug, secho, exceptions as exc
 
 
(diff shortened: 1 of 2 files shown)


wwitzel3:
👍

PR #145, opened 2016-05-10
Feature survey crud

jeremypruitt:
Currently you can not add surveys to Ansible Tower Job Templates with the tower-cli tool. This request is to extend the library (not the CLI tool) to create surveys on Job Templates.

It is ugly as it seems the library is opinionated around some things that made sense before Ansible released survey support in Tower. Most resources in Tower have top level endpoints, like /inventories, /organizations, /job_templates, /projects, etc. However, surveys do not exist at a top level like /surveys or /survey_specs. Instead they are a sub resource of Job Templates and their endpoint is namespaced by job template so the endpoint is like this: 

   /job_templates/5/survey_spec

Where "5" is the ID of the Job Template to which you are adding a survey.


jeremypruitt:
So sorry did not mean to open this here. :)

AlanCoding:
I know this PR might not have been intended, but I'll leave some notes anyway. CRUD operations for JT surveys are certainly one of the top features we are in need of. Very cool to see this. I would also like to see an ability to look up the JT by name, but this can be added in easily by use of the decorator without changing anything else. Let me know if there's any way I can help.

Issue #144, opened 2016-05-04
Can not print a resource's fields


AlanCoding:
This is relevant for anyone using tower-cli as a python library.

Create some resource, like `res = tower_cli.get_resource('foobar')`, then try to inspect `res.fields`. This should be a list, but an error is encountered half way through:

```
[<Field: name (str, unique)>, <Field: description (str, not required)>, *** AttributeError: 'function' object has no attribute 'replace'
```

This appears to be due to the fact that [our custom types](https://github.com/ansible/tower-cli/blob/master/lib/tower_cli/utils/types.py) don't have [a `__name__` attribute](https://github.com/ansible/tower-cli/blob/master/lib/tower_cli/models/fields.py#L71). We should to go through all our custom types and assure that `__name__` is there and that the `__repr__` function works.

It looks like our inherited `ParamType`s from `click` also don't have this stuff, so there may be some potential for contributing upstream.

jangsutsr:
All `__name__` attributes are added to custom types. However there are also external field types that do not have __name__  attribute (like click.Choice). So for now I use `type(self.type).__name__` to represent the name of all field type names. Not sure if it is okay though.

PR #148, opened 2016-05-31
Solve issue 'Can not print a resource's fields #144'

jangsutsr:


tests/test_utils_types.py
```diff
@@ -22,7 +22,10 @@
 from tower_cli.api import client
 from tower_cli.utils import exceptions as exc, types
 
-from tests.compat import unittest, mock
+try:
+    from tests.compat import unittest, mock
```
AlanCoding: This is kind of circular with the intent of `compat.py`. No other tests do it this way, and it would be better to keep the same format.

tests/test_utils_types.py
```diff
@@ -133,3 +136,23 @@ def test_get_metavar(self):
...
+class GeneralTests(unittest.TestCase):
+    """A set of tests that are not specific to any type.
+    """
+    def test_type_name(self):
+        """Establish that custom types are equipped with __name__ field and the
+        upstream click.Choice's issue of having no __name__ field is properly
+        dealt with.
+        """
+        from os import walk
+        from tower_cli import get_resource
```
AlanCoding: We would like to keep import statements at the top of the file. Also, below, that's a very inventive solution for grabbing all of the different resources, but I would rather see more consistency with the rest of the code base in order to loop through all types of resources.

lib/tower_cli/utils/types.py
```diff
@@ -65,7 +67,7 @@ class Related(click.types.ParamType):
     """A subclass of click.types.ParamType that represents a value
     related to another resource.
     """
-    name = 'related'
```
AlanCoding: I think that removing `name` is likely to cause problems because other things will assume that it exists.

PR #151, opened 2016-06-01
enhanced bug fix for issure #144

jangsutsr:


tests/test_utils_types.py
```diff
@@ -133,3 +134,30 @@ def test_get_metavar(self):
...
+            'job',
+            'job_template',
+            'organization',
+            'project',
+            'team',
+            'user'
+        ]
+        for res_name in names:
+            res = get_resource(res_name)
+            self.assertEqual(type(res.fields), type([]))
```
AlanCoding: One thing I would like to know - was this failing before the fix? I know it threw the attribute error, but if that happened in a try-except block, it could still return a list with the error text as an element. But if running the tests previously showed an error, then this is fine.

AlanCoding:
👍

Issue #147, opened 2016-05-18
Allow more formatting styles


AlanCoding:
As of now, we allow `human` and `json` for output styles specified by the `--format` option. The human option is a brief snippet of info in a table, while the json is everything that the server serializes.

It would be nice to have some in-between options. I would suggest 2 enhancement features:

 - ability to dump the output as YAML with the nice-looking flow style (easy)
 - ability to specify the fields to be printed with a list-valued option

Issue #119, opened 2016-01-12
Can't pass 'Instance Filters' parameter in tower-cli during create inventory group


ryanooi:
Can't pass 'Instance Filters' parameter in tower-cli during create inventory group

```
➜  ~  tower-cli version
Ansible Tower 2.4.3
Tower CLI 2.3.1 
```
---

```
➜  ~  tower-cli group create --help
Usage: tower-cli group create [OPTIONS]

  Create a group and, if necessary, modify the inventory source within the
  group.

Options:
  --name TEXT                     The name field.
  --description TEXT              The description field.
  --inventory INVENTORY           The inventory field.
  --variables FILENAME            The variables field.
  --credential CREDENTIAL         The cloud credential to use.
  --source [manual|ec2|rax|vmware|gce|azure|openstack]
                                  The source to use for this group.
  --source-regions TEXT           Regions for your cloud provider.
  --source-vars TEXT              Override variables found on source with
                                  variables defined in this field.
  --overwrite BOOLEAN             Delete child groups and hosts not found in
                                  source.
  --overwrite-vars BOOLEAN        Override vars in child groups and hosts with
                                  those from the external source.
  --update-on-launch BOOLEAN      Refresh inventory data from its source each
                                  time a job is run.
  --fail-on-found                 If used, return an error if a matching
                                  record already exists.  [default: False]
  --force-on-exists               If used, if a match is found on unique
                                  fields, other fields will be updated to the
                                  provided values. If False, a match causes
                                  the request to be a no-op.  [default: False]
  -h, --tower-host TEXT           The location of the Ansible Tower host.
                                  HTTPS is assumed as the protocol unless
                                  "http://" is explicitly provided. This will
                                  take precedence over a host provided to
                                  `tower config`, if any.
  -u, --tower-username TEXT       Username to use to authenticate to Ansible
                                  Tower. This will take precedence over a
                                  username provided to `tower config`, if any.
  -p, --tower-password TEXT       Password to use to authenticate to Ansible
                                  Tower. This will take precedence over a
                                  password provided to `tower config`, if any.
  -f, --format [human|json]       Output format. The "human" format is
                                  intended for humans reading output on the
                                  CLI; the "json" format is intended for
                                  scripts that wish to parse output. Note that
                                  the "json" format provides more data.
  -v, --verbose                   Show information about requests being made.
  --insecure                      Turn off insecure connection warnings. Set
                                  config verify_ssl to make this permanent.
  --help                          Show this message and exit.
[...]

PR #152, opened 2016-06-06
Solve issue #119 and add YAML support

jangsutsr:
Connect #119 

Connect #147


AlanCoding:
Nice. These are some good updates to have :+1:

PR #154, opened 2016-06-17
Enhance yaml output

jangsutsr:
Add extra parsing steps so that the output yaml will maintain the original key order

### DIFF
--- a/lib/tower_cli/models/base.py
+++ b/lib/tower_cli/models/base.py
@@ -41,6 +41,7 @@ from tower_cli.utils.command import Command
 from tower_cli.utils import debug, secho
 from tower_cli.utils.data_structures import OrderedDict
 from tower_cli.utils.decorators import command
+from tower_cli.utils.parser import ordered_dump
 
 
 class ResourceMeta(type):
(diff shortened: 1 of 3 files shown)


lib/tower_cli/models/base.py
```diff
@@ -312,9 +312,19 @@ def _format_yaml(self, payload):
                 """Convert the payload into a YAML string with proper
                 indentation and return it.
                 """
-                payload = json.loads(json.dumps(payload))
-                return yaml.safe_dump(payload, indent=2, allow_unicode=True,
-                                      default_flow_style=False)
+                def ordered_dump(data, stream=None, Dumper=yaml.Dumper, **kws):
```
AlanCoding: Putting this in a function seems a little more complicated than we need. We don't have any intention of reusing it, so maybe we could just remove the function and keep the last line as `return yaml.dump(...)`?

AlanCoding: Thinking about this some more.... I wouldn't mind `ordered_dump` being a function kept in a workspace for functions. This might help out later with other items on the agenda. Perhaps `lib/utils/parsers.py`.

AlanCoding:
👍  Nice work!

PR #155, opened 2016-06-20
Add RPM specfile and Makefile

ghjm:
This PR adds a preliminary RPM build process.

- Spec file by @wenottingham, lightly edited so it works
- Makefile by @ghjm, builds .tar.gz and el6 and el7 RPMs

There is a known issue with an unpackaged dependency on sorted-dict, and there are probably other issues as well. Little to no testing has been done yet.

### DIFF
--- a/packaging/rpm/ansible-tower-cli.spec
+++ b/packaging/rpm/ansible-tower-cli.spec
@@ -1,0 +1,53 @@
+%{!?python_sitelib: %define python_sitelib %(%{__python} -c "from distutils.sysconfig import get_python_lib; print get_python_lib()")}
+
+Name:           ansible-tower-cli
+Version:        2.3.1
+Release:        1%{?dist}
+Summary:        Commandline interface for Ansible Tower
+Group:          Development/Tools
+License:        ASL 2.0
+URL:            https://github.com/ansible/tower-cli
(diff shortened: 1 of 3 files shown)


jlaska:
:shipit:, nice addition, thanks @ghjm

ghjm:
Other than the dependency on sdict, this installs and apparently works on EL7. However, on EL6, it fails to install due to an unsatisfied dependency on python-colorama, which is present in EPEL7 but not EPEL6.

packaging/rpm/ansible-tower-cli.spec
```diff
@@ -0,0 +1,53 @@
+%{!?python_sitelib: %define python_sitelib %(%{__python} -c "from distutils.sysconfig import get_python_lib; print get_python_lib()")}
+
+Name:           ansible-tower-cli
+Version:        2.3.1
```
AlanCoding: note to self for later - we'll add this to the list of things to update pre-release

AlanCoding:
👍 Let's get this moving, and we'll give a good look at package dependency soon. Thanks!

packaging/rpm/ansible-tower-cli.spec
ghjm: I'll take a look at parsing the version number out from existing sources (like the VERSION file) to save you the trouble.

Issue #153, opened 2016-06-17
Error when syncing a group with --monitor


AlanCoding:
Run the command `tower-cli group sync "group_name" --monitor`, where group_name is the name of some group with a cloud source. Or, use the primary key in place of the group name.

When the job finishes and it tries to print the output, the traceback is produced:

```
Current status: successful
Traceback (most recent call last):
  File "/Users/alancoding/.virtualenvs/cli2/bin/tower-cli", line 4, in <module>
    __import__('pkg_resources').run_script('ansible-tower-cli==2.3.1', 'tower-cli')
  File "/Users/alancoding/.virtualenvs/cli2/lib/python2.7/site-packages/pkg_resources/__init__.py", line 724, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/Users/alancoding/.virtualenvs/cli2/lib/python2.7/site-packages/pkg_resources/__init__.py", line 1650, in run_script
    exec(code, namespace, namespace)
  File "/Users/alancoding/.virtualenvs/cli2/lib/python2.7/site-packages/ansible_tower_cli-2.3.1-py2.7.egg/EGG-INFO/scripts/tower-cli", line 80, in <module>
    cli()
  File "/Users/alancoding/.virtualenvs/cli2/lib/python2.7/site-packages/click/core.py", line 716, in __call__
    return self.main(*args, **kwargs)
  File "/Users/alancoding/.virtualenvs/cli2/lib/python2.7/site-packages/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File "/Users/alancoding/.virtualenvs/cli2/lib/python2.7/site-packages/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/alancoding/.virtualenvs/cli2/lib/python2.7/site-packages/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/alancoding/.virtualenvs/cli2/lib/python2.7/site-packages/click/core.py", line 889, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/alancoding/.virtualenvs/cli2/lib/python2.7/site-packages/click/core.py", line 534, in invoke
    return callback(*args, **kwargs)
[...]

PR #157, opened 2016-06-21
Solve issue #153

jangsutsr:
Change `__getitem__` to `get(item, default='N/A')` to handle exception of missing response fields. Now the output is.
```
(tower_cli_devel) sitan-OSX:tower-cli sitan$ tower-cli group sync 2 --monitor
== ======================= ========= 
id          name           inventory 
== ======================= ========= 
 8 many_hosts (many_hosts) N/A
== ======================= ========= 
```

*Update*: Note after further modifications, inventory can be properly shown now.
```
(tower_cli_devel) sitan-OSX:tower-cli sitan$ tower-cli group sync 3 --monitor
== ========= =========     
id   name    inventory 
== ========= ========= 
12 爷爷好 (叫爷爷)         3
== ========= ========= 

```

However, as presented. `_format_human` seems to have insufficient support on unicodes.

### DIFF
--- a/lib/tower_cli/models/base.py
+++ b/lib/tower_cli/models/base.py
@@ -368,7 +368,8 @@ class BaseResource(six.with_metaclass(ResourceMeta)):
                 for col in columns:
                     widths[col] = max(
                         len(col),
-                        *[len(six.text_type(i[col])) for i in raw_rows]
+                        *[len(six.text_type(i.get(col, 'N/A')))
+                          for i in raw_rows]
                     )
 
                 # It's possible that the column widths will exceed our terminal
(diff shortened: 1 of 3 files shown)


AlanCoding:
Could you post the output of the command to sync a group with the --monitor flag after this change?

AlanCoding:
https://github.com/ansible/tower-cli/blob/master/lib/tower_cli/models/base.py#L907

We should do a GET to the resource and include that data in the stuff returned from the `monitor` method. That can be combined with the status-type dictionary (I'm ambivalent to this). This will fill in an actual value for `inventory` in the above example, instead of just N/A.

AlanCoding:
👍

Issue #156, opened 2016-06-20
Remove sdict library


AlanCoding:
The requirement `dict.sorted` doesn't have a vendored rpm package associated with it, and it is the only one of our libraries like this. It seems that it's not required for any obvious reason, and we have an ordered dump of YAML in the works right now, which might serve our purposes whenever we need to preserve order.

This issue is to remove the package and work around any problems that come up doing so.

PR #161, opened 2016-06-24
Solve issue #156

jangsutsr:
Removed all occurrences of `sdict`. Now the outcome of `status` command becomes:
```
(tower_cli_devel) sitan-OSX:tower-cli sitan$ tower-cli project status 5
========== ====== ======= 
  status   failed elapsed 
========== ====== ======= 
successful  false   5.023
========== ====== ======= 
```

Note I did not modify `requirements.txt` here. Further modifications might be required.

### DIFF
--- a/lib/tower_cli/models/base.py
+++ b/lib/tower_cli/models/base.py
@@ -25,7 +25,6 @@ import re
 import sys
 import time
 from copy import copy
-from sdict import adict
 
 import six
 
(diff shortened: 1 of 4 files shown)


AlanCoding:
I've been testing this branch. I don't see any problems with the ordering. Lke:

```
== ====== ==== ======== 
id  name  kind username 
== ====== ==== ======== 
18 cred   ssh  pi
```

Here, id and name come first - as we would like. I'm curious if you saw any ordering mismatches. But anyway, looks good to me 👍

PR #162, opened 2016-06-24
Issue159

jangsutsr:
Any org association actions are meaningless with zero or one subject. So I made all of them required options.


AlanCoding:
In addition to `organization`, we have associate / dissociate commands on

 - `team`
 - `group`
 - `host`

It would be good to add this marker to those as well, although I just introduced the case of host with a recent push. If you want to do those as a separate PR, I give this a 👍

Issue #163, opened 2016-06-24
Adding a project that already exists (with the --monitor flag) indicates a change happened


AlanCoding:
If a project already exists and I issue a `tower-cli project create` command with `--name` targeting that project and the `--monitor` flag set too, the output is yellow. This is indicating that a change was made, but there was no change made. For all other resources, if I do this, the output is green, indicating that no change has occurred because the default force_on_exists is False.

PR #166, opened 2016-06-28
Solve issue #163

jangsutsr:
Apply a stricter if condition to solve #163. A side benefit is to avoid any unwarranted scm updates of existing projects when incorrectly calling `tower-cli project create` with monitor enabled.

### DIFF
--- a/lib/tower_cli/resources/project.py
+++ b/lib/tower_cli/resources/project.py
@@ -98,7 +98,7 @@ class Resource(models.Resource, models.MonitorableResource):
             org_resource._assoc('projects', org_pk, project_id)
 
         # if the monitor flag is set, wait for the SCM to update
-        if monitor:
+        if monitor and answer.get('changed', False):
             return self.monitor(project_id, timeout=timeout)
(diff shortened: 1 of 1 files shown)


lib/tower_cli/resources/project.py
```diff
@@ -98,7 +98,7 @@ def create(self, organization=None, monitor=False, timeout=None,
             org_resource._assoc('projects', org_pk, project_id)
 
         # if the monitor flag is set, wait for the SCM to update
-        if monitor:
+        if monitor and answer['changed']:
```
AlanCoding: The parent method should alway give a dictionary with `changed` in it (after having checked the code), but it would still be better to do ` and answer.get('changed', False)` here, in order to alleviate any of those concerns.

But that's my only comment. Nice work tracking this down!

jangsutsr: Agreed, never harm to be more careful. Changes made.

AlanCoding:
👍

PR #169, opened 2016-07-07
Juggle library versions to accommodate installs on RHEL6

AlanCoding:
I'm still looking into testing `python-importlib`, but I'm going to try to get a full install done based off of this branch.


AlanCoding:
Tested, and I believe this is working on EL6. It was working on EL7, and it might still be.

Issue #178, opened 2016-07-22
Find a better way of representing oversized debug log messages.


jangsutsr:
Some debug log messages are oversized due to clarification:
```
*** DETAILS: Checking for an existing record. *********************************
GET https://ec2-54-234-179-218.compute-1.amazonaws.com/api/v1/hosts/
Params: {'name': u'localhost'}

*** WARNING: Multiple Results found. Possibly caused by missing required options.
```

There are two solutions: separate long line or allow indefinite log length.

PR #179, opened 2016-07-24
Tentative new styled debug log message.

jangsutsr:
Connect to issue #178 .

### DIFF
--- a/lib/tower_cli/utils/debug.py
+++ b/lib/tower_cli/utils/debug.py
@@ -29,15 +29,32 @@ def log(s, header='', file=sys.stderr, nl=1, **kwargs):
     if not settings.verbose:
         return
 
-    # If this is a "header" line, make it a header.
+    # Construct resulting string to stderr.
+    word_arr = s.split(' ')
+    result = []
     if header:
-        s = '*** %s: %s %s' % \
-            (header.upper(), s, '*' * (72 - len(header) - len(s)))
+        word_arr.insert(0, '%s:' % header.upper())
+    i = 0
+    while i < len(word_arr):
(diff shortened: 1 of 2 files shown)


AlanCoding:
I'm not liking this much in its current form because of the impact that it is having on blank lines.

```
tower-cli credential create --name blank_credential2 --kind ssh --organization Default -v
*** DETAILS: Checking Project API Details. ************************************
*** OPTIONS http://192.168.100.128:8013/api/v1/credentials/ *******************
***  **************************************************************************
*** GET http://192.168.100.128:8013/api/v1/credentials/ ***********************
*** Params: {'organization': 1, 'kind': 'ssh', 'name': u'blank_credential2'} **
***  **************************************************************************
*** POST http://192.168.100.128:8013/api/v1/credentials/ **********************
*** Data: {'kind': 'ssh', 'name': u'blank_credential2', 'organization': 1} ****
***  **************************************************************************
Resource changed.
== ================= ==== ======== 
id       name        kind username 
== ================= ==== ======== 
16 blank_credential2 ssh
== ================= ==== ======== 
```

I think it's a good concept, but it needs to be narrowed to address only the very specific case of an oversized "details" message. Even the "Data:" and "Params:" would probably do best to avoid the asterisks.

jangsutsr:
@AlanCoding Thx for pointing out. Double checked to find I made a mistake: padding with asterisks *only* applies to messages with headers. Will correct it and add test cases.

jangsutsr:
@AlanCoding It should be working now: only messages with header will be padded.

AlanCoding:
:+1: works fine as far as I've tested it now

Issue #174, opened 2016-07-22
Add support for tower-3.0.0 notification templates


jlaska:

jangsutsr:
Features summary (these will be filed in unit test):
- [ ] Job template should be able to associate itself with an existing notification template.
```
(tower_cli_devel) sitan-OSX:tower-cli sitan$ tower-cli job_template associate_notification --job-template 5 --notification 3 -v -d --insecure
GET https://ec2-54-234-179-218.compute-1.amazonaws.com/api/v1/job_templates/5/notification_templates_any/
Params: {u'id': 3}

POST https://ec2-54-234-179-218.compute-1.amazonaws.com/api/v1/job_templates/5/notification_templates_any/
Data: {u'id': 3, u'associate': True}

OK. (changed: true)
```
- [ ] Job template should be able to disassociate itself from an associated notification template.
```
(tower_cli_devel) sitan-OSX:tower-cli sitan$ tower-cli job_template disassociate_notification --job-template 5 --notification 3 -v -d --insecure
GET https://ec2-54-234-179-218.compute-1.amazonaws.com/api/v1/job_templates/5/notification_templates_any/
Params: {u'id': 3}

POST https://ec2-54-234-179-218.compute-1.amazonaws.com/api/v1/job_templates/5/notification_templates_any/
Data: {u'id': 3, u'disassociate': True}

OK. (changed: true)
```
- [ ] Create a notification template in isolation.
```
(tower_cli_devel) sitan-OSX:tower-cli sitan$ tower-cli notification create --name foo --description bar --notification-type slack --channels a --channels b --token hey -v -d --insecure
*** DETAILS: Checking for an existing record. *********************************
GET https://ec2-54-234-179-218.compute-1.amazonaws.com/api/v1/notification_templates/
Params: {u'name': u'foo'}

*** DETAILS: Writing the record. **********************************************
POST https://ec2-54-234-179-218.compute-1.amazonaws.com/api/v1/notification_templates/
Data: {'notification_type': 'slack', 'notification_configuration': {'channels': (u'a', u'b'), 'token': u'hey'}, 'description': u'bar', 'name': u'foo'}

== ==== =========== ================= 
id name description notification_type 
== ==== =========== ================= 
13 foo  bar         slack
== ==== =========== =================
```
- [ ] Create a notification template under an existing job template.
```
[...]

PR #181, opened 2016-07-27
Implement Notification Templates

jangsutsr:
Connect to issue #174. See #174 for detailed specifications.

### DIFF
--- a/docs/NOTIFICATION_MANAGEMENT.md
+++ b/docs/NOTIFICATION_MANAGEMENT.md
@@ -1,0 +1,88 @@
+## Introduction - What Notification Templates Are
+
+Starting with Ansible Tower 3.0, users can create and associate
+notification templates to job templates. Whenever a job template undergoes
+a successful/unsuccessful job run, all its related notification templates
+will send out notifications to corresponding notification receiver, indicating
+the status of job run.
+
+Note although tower-cli uses `notification` as resource name, it actually refers
+to notification templates and related operations.
+ 
+## Managing Notification Templates with tower-cli
+
+To see the commands available for notification templates, see 
+`tower-cli notification --help`. Within a specific command, get the help 
+text with `tower-cli notification <command> --help`.
+
+The arguments for all notification commands follow the same pattern, although
+not all arguments are mandatory for all commands. The structure follows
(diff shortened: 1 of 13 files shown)


docs/NOTIFICATION_MANAGEMENT.md
```diff
@@ -0,0 +1,88 @@
...
+tower-cli notification create --name foo --description bar --notification-type slack --channels a --channels b --token hey --job-template 5
+
+# Get exactly one notification template.
+tower-cli notification get --name foo
+
+# Get a list of notification templates under certain criteria.
+tower-cli notification list --notification-type irc
+
+# Modify an existing notification.
+tower-cli notification modify --description foo --token hi 17
```
AlanCoding: Everything here is great, by the way. The only thing I notice as sticking out to me is the fact that `token` is a protected field, and if it is not given, the modification gets sent with `$encrypted$` as its value. That's a fundamental limitation of the server, and there's literally no way we can let the user modify anything without them also giving a new token.

So the only thing I could see being done about it would be to raise an error if the token is not provided in the`modify` command. This might be complicated if the encrypted fields are specific to the type. In that case, we could probably leave it as a TODO for later.

jangsutsr: Fixed. Now encrypted fields become mandatory and failure to providing them during modification would trigger exception.

lib/tower_cli/resources/credential.py
```diff
@@ -144,6 +144,8 @@ class Resource(models.Resource):
     vault_password = models.Field(password=True, required=False)
 
     def create(self, **kwargs):
+        """Create a credential.
+        """
```
AlanCoding: This is a good catch. I was vaguely aware that we were missing help text in some places, but I had not got around to making an issue for it.

jangsutsr: Indeed. I have swept through the help texts and this missing issue should be solved altogether after merging.

AlanCoding:
I give this my 👍 . This is an impressive contribution. There are many complex faucets that must be handled for this resource, and they all appear to be covered.

Issue #172, opened 2016-07-12
Allow passing any needed prompted items to job launch


wenottingham:
For anything in Tower that allows prompting at runtime, tower-cli should allow you to specify it when launching a job.

jangsutsr:
Checked to find the machine credential-related passwords are not implemented yet. Moving back to 'devel'.

PR #182, opened 2016-07-27
Modify credential field help text to indicate prompt-at-run-time.

jangsutsr:
Connect to #172. Luckily all the prompt functionalities are implemented in job launch. Just need to be more specific in the help text.

### DIFF
--- a/lib/tower_cli/resources/credential.py
+++ b/lib/tower_cli/resources/credential.py
@@ -20,6 +20,9 @@ from tower_cli.utils import types, debug
 from tower_cli.api import client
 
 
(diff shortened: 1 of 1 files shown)


AlanCoding:
This looks good, if I understand correctly, but I just want to make sure I do. The intended use case here is that we provide `--ssh-key-unlock=ASK`, for example. Then when the user does that, the server interprets this as a signal to prompt on launch. Is that correct?

jangsutsr:
@AlanCoding Exactly, I have verified that.

AlanCoding:
👍

PR #195, opened 2016-08-08
Bundle some commands into Makefile to ease devel installation.

jangsutsr:
Now typing `make install` would do both clean-up and installation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,4 +42,7 @@ rpm-build/.exists:
 	mkdir -p rpm-build
 	touch rpm-build/.exists
 
-
+# For devel convinience
+install:
+	sudo rm -rf dist/ build/ 
+	sudo python setup.py install


Makefile
```diff
@@ -42,4 +42,7 @@ rpm-build/.exists:
 	mkdir -p rpm-build
 	touch rpm-build/.exists
 
-
+# For devel convinience
+install:
+	sudo rm -rf dist/ build/
```
AlanCoding: To be more DRY, you could reuse the `clean` command (also add one more folder to its list to delete) and then do the install after that. I could see arguments both for and against. That's just something for you to consider, 👍  to this PR either way.

jangsutsr: Agreed. But since it is just for developer's convenience, let us leave it as it is for now.

Issue #198, opened 2016-08-10
Implement Schedules


jangsutsr:
Features summary:

- [x] No more than one unified job template fields should be provided.
```
(tower-cli-devel) dyn-160-39-203-38:~ siruitan$ tower-cli schedule get --job-template 1 --project 2 -v
Error: More than one unified job template fields provided, please tighten your criteria.
```
- [x] At least one unified job template field should be provided during creation.
```
(tower-cli-devel) dyn-160-39-203-38:tower-cli siruitan$ tower-cli schedule create --name hehe -v
Error: You must provide exactly one unified job template field during creation.
```
- [x] Click-related extra function attributes are passed on to decorated methods.
- [x] Docstring of the original method are passed on to decorated counterpart.
- [x] Correct endpoint is invoked according to unified jt field provided during creation.
```
(tower-cli-devel) dyn-160-39-203-38:tower-cli siruitan$ tower-cli schedule create --name hehe --job-template 5 --rrule "DTSTART:20160812T200122Z RRULE:FREQ=MONTHLY;INTERVAL=1;BYMONTHDAY=1" -v
*** DETAILS: Checking for an existing record. *********************************
GET http://192.168.99.100:8013/api/v1/job_templates/5/schedules/
Params: {u'name': u'hehe'}

*** DETAILS: Writing the record. **********************************************
POST http://192.168.99.100:8013/api/v1/job_templates/5/schedules/
Data: {'name': u'hehe', 'enabled': True, 'unified_job_template': 5, 'rrule': u'DTSTART:20160812T200122Z RRULE:FREQ=MONTHLY;INTERVAL=1;BYMONTHDAY=1'}

Resource changed.
== ==== ============ ================ ======= ======= 
id name job_template inventory_source project enabled 
== ==== ============ ================ ======= ======= 
 4 hehe N/A          N/A              N/A        true
== ==== ============ ================ ======= ======= 
```

PR #199, opened 2016-08-18
Implement schedules

jangsutsr:
Connect to #198.

### DIFF
--- a/lib/tower_cli/resources/schedule.py
+++ b/lib/tower_cli/resources/schedule.py
@@ -111,6 +111,11 @@ class Resource(models.Resource):
                            display=False)
 
     # Schedule-specific fields.
+    unified_job_template = models.Field(required=False, type=int,
+                                        help_text='Integer used to display'
+                                        ' unified job template in result, '
+                                        'Please don\'t use it for create/'
+                                        'modify.')
     enabled = models.Field(required=False, type=click.BOOL, default=True,
                            help_text='Whether this schedule will be used',
                            show_default=True)
--- a/lib/tower_cli/resources/schedule.py
+++ b/lib/tower_cli/resources/schedule.py
@@ -116,7 +116,7 @@ class Resource(models.Resource):
     rrule = models.Field(required=False, display=False,
                          help_text='Schedule rules specifications which is'
                          ' less than 255 characters.')
-    extra_data = models.Field(type=models.File('r', lazy=True), required=False,
+    extra_data = models.Field(type=types.Variables(), required=False,
                               display=False, help_text='Extra data for '
                               'schedule rules in the form of a .json file.')
 
--- a/lib/tower_cli/resources/schedule.py
+++ b/lib/tower_cli/resources/schedule.py
@@ -43,7 +43,7 @@ def jt_aggregate(func, is_create=False, has_pk=False):
         exc.UsageError: Either more than one unified jt fields are
             provided, or none is provided when is_create flag is set.
     """
-    def helper(kwargs):
+    def helper(kwargs, obj):
         """The helper function preceding actual function that aggregates
         unified jt fields.
         """
@@ -60,6 +60,7 @@ def jt_aggregate(func, is_create=False, has_pk=False):
                     )
         if unified_job_template is not None:
             kwargs['unified_job_template'] = unified_job_template[1]
+            obj.identity = tuple(list(obj.identity) + ['unified_job_template'])
             return '/'.join([UNIFIED_JT[unified_job_template[0]],
                              str(unified_job_template[1]), 'schedules/'])
         elif is_create:
@@ -68,7 +69,7 @@ def jt_aggregate(func, is_create=False, has_pk=False):
 
     def decorator_without_pk(obj, *args, **kwargs):
         old_endpoint = obj.endpoint
-        new_endpoint = helper(kwargs)
+        new_endpoint = helper(kwargs, obj)
         if is_create:
             obj.endpoint = new_endpoint
         result = func(obj, *args, **kwargs)
@@ -77,7 +78,7 @@ def jt_aggregate(func, is_create=False, has_pk=False):
 
     def decorator_with_pk(obj, pk=None, *args, **kwargs):
         old_endpoint = obj.endpoint
-        new_endpoint = helper(kwargs)
+        new_endpoint = helper(kwargs, obj)
         if is_create:
             obj.endpoint = new_endpoint
         result = func(obj, pk=pk, *args, **kwargs)
--- a/lib/tower_cli/resources/schedule.py
+++ b/lib/tower_cli/resources/schedule.py
@@ -1,0 +1,124 @@
+# Copyright 2016, Ansible by Red Hat.
+# Aaron Tan <sitan@redhat.com>
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import click
+
+from tower_cli import models
+from tower_cli.utils import types, exceptions as exc
+
+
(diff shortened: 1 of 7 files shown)


michelleperz:
@jangsutsr We just had a customer put in a ticket who would like the ability to modify the schedules of management jobs using tower-cli. These don't show up in the normal job template list  so would need to be tied to unified job templates instead.

jangsutsr:
@michelleperz Thx for pointing out! Will implement system management job to work with schedule.

AlanCoding:
I think the challenge with allowing for system JT schedules is making the switch of the base endpoint. Whereas we have `/api/v1/job_templates/N/`, we have `/api/v1/system_job_templates/N/` for system JTs. This might be simple to over-ride with the logic you have here. It does seem a little inconsistent because I don't think we can manage system JTs to begin with in tower-cli. If that starts to scope-creep, let's make a new issue for adding support for them generally.

lib/tower_cli/resources/schedule.py
```diff
@@ -0,0 +1,125 @@
...
+    # Unified jt fields. note these fields will only be used during creation.
+    # Plus, one and only one field should be provided.
+    job_template = models.Field(type=types.Related('job_template'),
+                                required=False)
+    inventory_source = models.Field(type=types.Related('inventory_source'),
+                                    required=False)
+    project = models.Field(type=types.Related('project'), required=False)
+
+    # Schedule-specific fields.
+    enabled = models.Field(required=False, type=click.BOOL, default=True,
```
AlanCoding: I would like the default to be shown in the help text. Experimenting with it, it isn't, but I don't know why. As a hack, you could even add it manually to the help text.

lib/tower_cli/resources/schedule.py
```diff
@@ -0,0 +1,125 @@
...
+                           help_text='Whether this schedule will be used')
+    rrule = models.Field(required=False, display=False,
+                         help_text='Schedule rules specifications which is'
+                         ' less than 255 characters.')
+    extra_data = models.Field(type=models.File('r', lazy=True), required=False,
+                              display=False, help_text='Extra data for '
+                              'schedule rules in the form of a .json file.')
+
+Resource.create = jt_aggregate(Resource.create, is_create=True)
```
AlanCoding: It seems like it would work to do

```python
Resource.create = resources.command(jt_aggregate(Resource.create, is_create=True))
```

as opposed to putting the `value._cli_command_attrs.setdefault(attkey, attval)` in a try-except block.

lib/tower_cli/resources/schedule.py
```diff
@@ -0,0 +1,125 @@
...
+
+    return decorator
+
+
+class Resource(models.Resource):
+    cli_help = 'Manage schedules within Ansible Tower.'
+    endpoint = '/schedules/'
+
+    # General fields.
+    name = models.Field(unique=True)
```
AlanCoding: I think this should either be put in the `identity` tuple or marked as required (probably the former). There's no indication that this needs to be provided, but it won't work without it.

AlanCoding:
I don't have a problems with the columns that you're trying to provide, but they don't seem to work. For example, in the following example, schedule pk=5 is a schedule on a job template.

```
$ tower-cli schedule get 5
== ==== ============ ================ ======= ======= 
id name job_template inventory_source project enabled 
== ==== ============ ================ ======= ======= 
 5 asdf N/A          N/A              N/A        true
== ==== ============ ================ ======= ======= 
```

It should figure out that it is of the "job_template" type and then put the pk in that column.

Alternatively, you might have better luck just giving the unified_job_template pk. This won't conflict because of the nature of polymorphic models. However, I think it would still be good to provide the resource type as another column, if you do this.

This constant translation of input arguments to a different internal value is the same magnitude of task that I was doing with roles earlier.

lib/tower_cli/resources/schedule.py
jangsutsr: Looks like it is controlled by Click's `show_default` kwarg, I didn't add that but it turns out even adding it won't generate default hints. Guess it has something to do with the decorator, will dig into it.

lib/tower_cli/resources/schedule.py
jangsutsr: Just removed that try-except block and everything worked fine. I am not even sure the reason for adding it, maybe some other mechanisms just solved the original issue.

lib/tower_cli/resources/schedule.py
jangsutsr: Marking as required is not needed here as Click convention sets `required`'s default value to `True`.

jangsutsr:
@AlanCoding Hidding display of unified-job-related fields for now. Adding UJT pk is a good idea, while providing resource types an another column could be a bit tricky to implement, as it is not part of the model fields and extra logics are needed to extract resource type from return values, but I will try.

lib/tower_cli/resources/schedule.py
AlanCoding: ok, it's showing correctly now testing it now.

lib/tower_cli/resources/schedule.py
```diff
@@ -0,0 +1,127 @@
...
+                           display=False)
+
+    # Schedule-specific fields.
+    enabled = models.Field(required=False, type=click.BOOL, default=True,
+                           help_text='Whether this schedule will be used',
+                           show_default=True)
+    rrule = models.Field(required=False, display=False,
+                         help_text='Schedule rules specifications which is'
+                         ' less than 255 characters.')
+    extra_data = models.Field(type=models.File('r', lazy=True), required=False,
```
AlanCoding: Could we change this to the Variables type that I've used in some other models? That would let them use either a file (distinguished by the `@` starting character) or direct CLI arguments.

lib/tower_cli/resources/schedule.py
AlanCoding: Maybe the tests are working now, but I just checked out your branch and got the `AttributeError: 'function' object has no attribute '_cli_command_attrs'` error. This was after installing and then trying a create command.

jangsutsr: Find the cause: some inherited methods are not decorated with `resources.command`. They would work fine when called directly, but introduce traceback when being imported by related fields.

The latest commit should have fixed all of them.

AlanCoding: I'm still getting that same error after picking up your latest commit. I've removed the egg.info files and the build folder and still no luck.

lib/tower_cli/resources/schedule.py
jangsutsr: Okay, I figured that out: `list` and `get` are not showing default value because default values are muted there by setting `ignore_defaults` to `True`, The other methods work fine.

lib/tower_cli/resources/schedule.py
jangsutsr: Varible type added.

lib/tower_cli/resources/schedule.py
jangsutsr: Caught the one missing and squashed it.

AlanCoding:
Checked out recent. Comments here.

The job template column is gone. If we could somehow provide `unified_job_template` column, that would still be very helpful. I guess not required for MVP right now.

```
$ tower-cli schedule get --name "quack regularily" -v
GET http://server/api/v1/schedules/
Params: {'name': u'quack regularily'}

== ================ ======= 
id       name       enabled 
== ================ ======= 
 4 quack regularily    true
== ================ ======= 
```

The `modify` command is putting in defaults that it really shouldn't be. In the following example, it set `enabled` to True. I can repeat this where I had `enabled` set to False before that command, and it changes it back. That's definitely a bug.

```
$ tower-cli schedule modify --name asdf --rrule "DTSTART:20160930T131334Z RRULE:FREQ=WEEKLY;INTERVAL=1;BYDAY=SU" -v
GET http://server/api/v1/schedules/
Params: {u'name': u'asdf'}

PATCH http://server/api/v1/schedules/5/
Data: {'enabled': True, 'rrule': u'DTSTART:20160930T131334Z RRULE:FREQ=WEEKLY;INTERVAL=1;BYDAY=SU', 'name': u'asdf'}

Resource changed.
== ==== ======= 
id name enabled 
== ==== ======= 
 5 asdf    true
== ==== ======= 
```

Those are the only 2 points I have. The create, delete, get, and list commands are all working as they should be.

jangsutsr:
@AlanCoding That bug was caused by another decorator inaccuracy in `base.py`. Kind of surprised that it has been there for so long.

Issue #202, opened 2016-08-25
Display `ignored_fields` if job template launch returns it.


jangsutsr:
In Tower, the return value of job template launch (POST to `/api/v1/job_templates/\d/launch`) will sometimes return a dict named `ignored_fields` showing which data provided are ignored during job launch. We should include this as a warning in verbose mode as it can be proven valuable in debugging.

PR #203, opened 2016-08-26
Add `ignored_fields` display to job launch in verbose mode

jangsutsr:
Connect #202.

### DIFF
--- a/lib/tower_cli/resources/job.py
+++ b/lib/tower_cli/resources/job.py
@@ -212,8 +212,8 @@ class Resource(models.ExeResource):
         # If 'ignored_fields' is present in the returning json, display it
         # in verbose mode.
         if 'ignored_fields' in job_started.json():
-            debug.log('Some provided fields are ignored on the server side:',
-                      header='warning')
+            debug.log('List of ignored fields on the server side:',
+                      header='detail')
             for key, value in job_started.json()['ignored_fields'].items():
                 debug.log('{0}: {1}'.format(key, value))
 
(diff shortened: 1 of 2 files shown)


lib/tower_cli/resources/job.py
```diff
@@ -209,6 +209,14 @@ def launch(self, job_template=None, monitor=False, timeout=None,
         if supports_job_template_launch and not use_job_endpoint:
             job_id = job_started.json()['job']
 
+        # If 'ignored_fields' is present in the returning json, display it
+        # in verbose mode.
+        if 'ignored_fields' in job_started.json():
```
AlanCoding: This field will be present in the response even if no fields are ignored. As a specific example, if no variables are rejected, it will look like this:

```json
    "ignored_fields": {
        "extra_vars": {}
    },
```

It's these cases that the `in` logic will often miss. I realize that this might be very tricky for you to correctly check for.

We also can't _count_ on the fields being there, because it really won't be present in 2.4.5, which we still have to support. Let me know if you have any questions on these stipulations.

jangsutsr: @AlanCoding Agreed. IMO it is not worthwhile enough for adding extra logic to slow down the process here, My workaround is changing the prompt text: Make it a detail message for user to consult, rather than a warning that explicitly draws user's attention.

AlanCoding: The wording is fine, but I'm still not a fan of this implementation. Launching jobs are presumably the most frequently used part of tower-cli, and this is noise in the console output that shouldn't be necessary. If it's a normal job and everything is working fine, then I would opt to not show the message. It is more complicated to do it, but the logic can be added in the one-line conditional that you already have to deal with both 2.4.5 and 3.0 cases. i.e. `if 'ignored_fields' in job_started.json() and 'extra_vars' in job_started.json()['ignored_fields'] and job_started.json()['ignored_fields']['extra_vars']:`. That's not best-practice (you might want to store the JSON-ified version before that line), but the basic logic isn't unreasonable.

Issue #205, opened 2016-09-21
Test failure


Fale:
Trying to run tests on Fedora 26 with Python 3.5, I receive the following errors.
I'm not sure if it's a missing piece in my side or not.
On the same environment, running the tests against Python 2.7 is a success.

```
======================================================================
ERROR: Establish that if connection fails and protocol not given,
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/builddir/build/BUILD/tower-cli-3.0.0/tests/test_api.py", line 199, in test_failed_suggestion_protocol
    secho.assert_called()
  File "/usr/lib64/python3.5/unittest/mock.py", line 583, in __getattr__
    raise AttributeError(name)
AttributeError: assert_called

======================================================================
FAIL: Establish that if `tower-cli config` is sent an invalid key,
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/builddir/build/BUILD/tower-cli-3.0.0/tests/test_commands_config.py", line 193, in test_error_invalid_key
    self.assertEqual(result.exit_code, 1)
AssertionError: -1 != 1

======================================================================
FAIL: Establish that if no global config file exists, that tower-cli
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/builddir/build/BUILD/tower-cli-3.0.0/tests/test_commands_config.py", line 214, in test_error_no_global_config_file
    isdir.assert_called_once_with('/etc/tower/')
  File "/usr/lib64/python3.5/unittest/mock.py", line 802, in assert_called_once_with
    raise AssertionError(msg)
AssertionError: Expected 'isdir' to be called once. Called 0 times.

======================================================================
FAIL: Establish that if `tower-cli config` is called with both a value
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/builddir/build/BUILD/tower-cli-3.0.0/tests/test_commands_config.py", line 202, in test_error_value_and_unset
    self.assertEqual(result.exit_code, 2)
AssertionError: -1 != 2

======================================================================
FAIL: Establish that if we are given a key and no value, that the
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/builddir/build/BUILD/tower-cli-3.0.0/tests/test_commands_config.py", line 59, in test_key_and_no_value
    self.assertEqual(result.exit_code, 0)
AssertionError: -1 != 0

======================================================================
FAIL: Establish that if `tower-cli config` is called with no arguments,
----------------------------------------------------------------------
[...]

PR #210, opened 2016-09-27
Use documented test

Fale:
the assert_called() property is not documented and seems to be absent in py3.5


Fale:
This impact partialli on #205

PR #211, opened 2016-09-27
Add nightly (3.6) and accept if it fails

Fale:


AlanCoding:
👍  Looks like the nightly actually passed where 3.5 failed, but that's probably meaningful information about 3.6 changes. This will help alert us of incompatibilities before users encounter them.

PR #212, opened 2016-09-28
add --ask-skip-tags-on-launch

AlanCoding:


AlanCoding:
This isn't enough. We have to add it to some processing in job launch.

Issue #213, opened 2016-09-28
Standard Out Streaming


AlanCoding:
Details of the currently proposed spec on the table:
- [x] replace action of `--monitor` flag with standard out streaming
- [x] move current monitor action to a `--wait` flag for CI/CD environments that don't want console noise
- [x] during the pending phase, show the same thing that we currently do (i.e. call the "wait" method)
- [x] modularize the GET request itself, allow use as its own command, see #74 
- [x] tests

PR #220, opened 2016-10-06
Introduce Standard Out Streaming with wait as old monitoring

AlanCoding:
Connect #74 

Connect #213 
## Examples of Using This Feature

Launch a job from job template number 7. Stream its standard output.

`tower-cli job launch -J 7 --monitor`

Launch that job, hold on until it finishes, but use the old wait text (which will delete itself).

`tower-cli job launch -J 7 --wait`

Print the standard out of an existing job to the command line.

`tower-cli job stdout 119`

Sync an inventory group. Stream the standard out input from the inventory module.

`tower-cli group sync --name open_stack --monitor`

Note that you can grab both `status` and `stdout` for a project. This should raise eyebrows, because technically a project doesn't have standard out (only its project update does). Actually, we grab the last job ran and show the output from that. That design decision was made a while ago, but this PR extends the logic to connect more things.

Create a project and update it. Also stream the standard out of the update.

`tower-cli project create --name="AlanCoding examples" --organization "Default" --scm-type git --scm-url https://github.com/AlanCoding/permission-testing-playbooks.git --monitor`
## Game Plans

This contains a departure from previous behavior. Anyone who uses tower-cli to connect Jenkins to Tower would be best-advised to switch from `--monitor` flags to `--wait`. On the other hand, if you like details, don't change anything.

Those above examples should also probably be converted into docs sometime and somewhere.

As a final note, this is kind of a half-way toward giving tower-cli decent abstraction for Unified Job Templates versus Unified Jobs. I foresee this being built on top of if, for instance, there was a request to incorporate a new type of Unified Job Template that also had to manage its own unique type of jobs.

Additionally, a great amount of credit goes to @chrismeyersfsu for the original prototype of this.

### DIFF
--- a/tests/test_models_unified_jobs.py
+++ b/tests/test_models_unified_jobs.py
@@ -62,7 +62,7 @@ class StandardOutTests(unittest.TestCase):
                     lookup.return_value = 'foobar'
                     result = self.res.stdout(42)
                     assert not result['changed']
-                    mock_echo.assert_called_once_with('foobar', nl=0)
+                    mock_echo.assert_called_once_with('foobar', nl=1)
 
     def test_stdout_with_lookup(self):
         "Test that unified job will be automatically looked up."
--- a/tests/test_models_unified_jobs.py
+++ b/tests/test_models_unified_jobs.py
@@ -1,0 +1,96 @@
+# Copyright 2016, Ansible by Red Hat
+# Alan Rominger <arominge@redhat.com>
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import time
+import click
+
+import tower_cli
+from tower_cli.api import client
+
+from tests.compat import unittest, mock
+
+
+def project_update_registration(t):
+    t.register_json('/project_updates/54/', {
+        'elapsed': 1335024000.0,
+        'failed': False,
+        'status': 'successful',
+    }, method='GET')
+
+
+class StandardOutTests(unittest.TestCase):
+    """
+    Test that standard out lookup and wrapper methods
+    """
+    def setUp(self):
+        # Representative of a Unified Job Template
+        self.res = tower_cli.get_resource('project')
+        # Representative of a Unified Job
+        self.job_res = tower_cli.get_resource('job')
+
+    def test_lookup_stdout(self):
+        "Test the method that makes a call to get standard out."
+        with client.test_mode as t:
+            # note:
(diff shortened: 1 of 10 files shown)


lib/tower_cli/models/base.py
```diff
@@ -829,11 +830,158 @@ class MonitorableResource(ResourceMethods):
...
+        # resource is Unified Job, but pk not given
+        elif not pk:
+            unified_job = self.get(**kwargs)
+            pk = unified_job['id']
+
+        content = self.lookup_stdout(pk, start_line, end_line)
+        if len(content) > 0:
+            click.echo(content, nl=0)
+
+        return {"changed": False}
```
jangsutsr: Is it necessary to provide this return value since it doesn't actually contain any info? Because the job stdout might not end with newline and something like this occurs:
```
(tower_cli_devel) dyn-160-39-224-60:tower-cli sitan$ tower-cli job stdout 11
Waiting for results...OK. (changed: false)
```

AlanCoding: Getting the standard out of a pending/waiting job is a case I had just not considered. Valid point... I don't think we can avoid returning the dictionary without changing the context that this runs in. I would prefer to fix it within this method, maybe just add another line if it doesn't end in a line break.

lib/tower_cli/models/base.py
```diff
@@ -829,11 +830,158 @@ class MonitorableResource(ResourceMethods):
...
+            # Put the process to sleep briefly.
+            time.sleep(interval)
+
+            # Make request to get standard out
+            content = self.lookup_stdout(pk, start_line)
+
+            # In the first moments of running the job, the standard out
+            # may not be available yet
+            if not content.startswith("Waiting for results"):
+                line_count = content.count('\n')
```
jangsutsr: Can we rely on `\n` here? Is it possible that some task stdout would also include `\n`?

AlanCoding: The meta-code we would want is probably something like `count all "\n" which are not also a part of an escaped line break`. But the problem with this is that I can't figure out a playbook which would actually print the sequence of characters of "\n", and my attempts at escaping it so-far just fail the playbook. If you can come up with a playbook that prints it, then I would be happy to try it and build a test-case around it.

jangsutsr: I'll try finding some. In the mean time, strings are equipped with a built-in method `splitlines` that could be helpful.

lib/tower_cli/models/base.py
```diff
@@ -829,11 +830,158 @@ class MonitorableResource(ResourceMethods):
...
+
+        click.echo('\033[0;91m------Starting Standard Out Stream------\033[0m',
+                   nl=2, file=outfile)
+
+        # Poll the Ansible Tower instance for status and content,
+        # and print standard out to the out file
+        while result['status'] != 'successful':
+
+            # Polling loop exit conditions
+            if result['failed']:
```
AlanCoding: Good point. This is a holdover from the old method. I'll move this to after the standard out call and before the results call.

lib/tower_cli/models/base.py
AlanCoding: Your `splitlines` suggestion solved the problem. Unfortunately, I can't perfectly explain what situation was happening with the first line of a job standard out, but I've tried it against various values for `interval` and they all format correctly and the code is much more compact now.

PR #223, opened 2016-10-12
Remove rpm file and build file for EL6 and EL7 since this software is…

Fale:
… being officially packaged in EPEL.
If other repos will want the same package, it's probably better if they will get the one in EPEL, since is far more complete than the one in this repo.

https://bodhi.fedoraproject.org/updates/?packages=python-ansible-tower-cli


wenottingham:
? There are more RPM platforms than just EPEL, it's fairly traditional to have an upstream spec and a downstream spec.

Fale:
True, but both files seems to focus especially on EL6/7

AlanCoding:
Thanks for your contributions. We would like to maintain this ability to build rpm files, so I'm closing this out. We would welcome any improvements to our spec file in a new PR.

Issue #217, opened 2016-10-01
Figure out how to deal with numeric resource names


jangsutsr:
Right now tower-cli is not smart enough to know it's not dealing with a primary key. Maybe we can figure out a way.

AlanCoding:
This is going to be tough to accomplish. Our related fields are only equipped to deal with one parameter, and I don't intuitively expect type checking to work.

Another related concern are resources with duplicate names allowed. Let's say you want to specify a certain host within a certain inventory (otherwise multiple results will be found). Would we be able to optionally enable a multi-parameter input to the flag to deal with all these cases? I'm not sure.

PR #224, opened 2016-10-19
allow skiping digits conversion when associating hosts to groups

evgeni:
It is perfectly fine to have hosts, groups and other resources to be called with an all-digits name in Tower/Ansible.
However this can confuse `tower-cli` when one tries to pass such a "name" to it.

```
% tower-cli host associate --host client123 --group 1337

POST https://192.168.121.199/api/v1/hosts/2/groups/
Params: None
Data: {"associate": true, "id": 1337}

Response: {"detail":"Group matching query does not exist."}

% tower-cli host associate --host 4713 --group foo        
Error: The requested object could not be found.
```

This PR adds a way to skip this conversion in `types.Related` which is used for the lookup of the values. And uses this new feature to implement new parameters `--host-name` and `--group-name` to `tower-cli host (dis)associate`. There are probably more places where this will be useful, but I am currently only scratching my own itch.

### DIFF
--- a/lib/tower_cli/resources/host.py
+++ b/lib/tower_cli/resources/host.py
@@ -35,15 +35,31 @@ class Resource(models.Resource):
     @resources.command(use_fields_as_options=False)
     @click.option('--host', type=types.Related('host'))
     @click.option('--group', type=types.Related('group'))
-    def associate(self, host, group):
+    @click.option('--host-name', type=types.Related('host',
+                  convert_digits=False))
+    @click.option('--group-name', type=types.Related('group',
+                  convert_digits=False))
+    def associate(self, host, group, host_name, group_name):
         """Associate a group with this host."""
+        if host_name and not host:
+            host = host_name
+        if group_name and not group:
+            group = group_name
         return self._assoc('groups', host, group)
 
     @resources.command(use_fields_as_options=False)
     @click.option('--host', type=types.Related('host'))
     @click.option('--group', type=types.Related('group'))
-    def disassociate(self, host, group):
+    @click.option('--host-name', type=types.Related('host',
+                  convert_digits=False))
+    @click.option('--group-name', type=types.Related('group',
+                  convert_digits=False))
+    def disassociate(self, host, group, host_name, group_name):
         """Disassociate a group from this host."""
+        if host_name and not host:
+            host = host_name
+        if group_name and not group:
+            group = group_name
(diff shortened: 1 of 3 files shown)


AlanCoding:
Connect #217

AlanCoding:
This works, but it's a little bit of a one-off scenario. For this fix, I would prefer to do something inside of the `Related` type class and avoid adding more to the commands themselves. Just to throw out a suggestion, what if we allowed a leading special character to denote a name? For instance, consider this pattern:

`tower-cli host associate --host client123 --group "#1337"`

In this case, we could have the type class drop the leading `#` in the group parameter, and then look it up with "1337" as the name.

evgeni:
But wouldn't that just move the problem to resources actually called `#1337`? That's less probable, but still somehow wrong, IMHO.
Other tools (e.g. `hammer` from Foreman) usually just have multiple `--selector`s denoting different selection columns. Like `--host-id`, `--host-name` etc. No guessing is done at all.

AlanCoding:
It would shift the problem as you say, but at least it wouldn't make the action impossible, since you could still specify `--group ##1337`, although I see the reasons this is also not ideal. That's certainly a documentation challenge.

You have a good argument for the dual flags. At some point, we could also automatically apply two flags for all the related resource flags. That will clutter the help text some more, and it's much more difficult to globally implement. It's a decision I would like to solicit input from more people for.

AlanCoding:
Just to check in - this is still on the agenda. I'm considering an approach where we accept this, and then develop a method to automatically apply both the name & id flags to all many-to-many relationships. This can't be done exclusively in the `types.py`, but will involve laying down a lot of scaffolding in `base.py`. That's fine, and it could keep things DRY and support the use-pattern in the long-run.

evgeni:
Cool. Shoot if you need anything from my side please.

AlanCoding:
Here's a basic idea of how I see this could possibly be implemented:

https://github.com/ansible/tower-cli/compare/master...AlanCoding:options_by_name?expand=1

However, this is far from working, because it passes the arguments in as `project_name` if there is a related field `project`. So this feature would involve adding additional processing to check either/or the normal parameter or the named parameter.

Maybe there's another more elegant way to use click to allow multiple CLI aliases to the same function param, which we could process in utils.types.

This would be a pretty big feature and I don't have a technical direction that I can see working right now.

PR #233, opened 2016-12-08
don't mark the inventory_script resource as internal

evgeni:
that way it shows up in help etc
the resource as such is perfectly working from the command line, no need to hide it :)


AlanCoding:
Right, because someone might want to upload a new script. @jangsutsr was there any non-obvious reason to make this abstract? We might want to eventually patch test an patch up its functionality. I would certainly like to be able to upload a script from a file, for instance.

evgeni:
@AlanCoding you actually can, nothing forbids you to call `tower-cli inventory_script create …` right now. The flag is only supressing the inclusion in `--help` etc

evgeni:
ping @jangsutsr, any objections?

PR #234, opened 2016-12-08
allow provisioning callbacks when creating a new job template

evgeni:


evgeni:
ping, is there anything more needed before this can be merged?

PR #235, opened 2016-12-08
don't mark the job-type as required, we default to "run" if it's not set

evgeni:


evgeni:
ping, is there anything more needed before this can be merged?

AlanCoding:
Oh sorry, I think I misunderstood that it would give the person who raised the PR access to merge.

evgeni:
I think one needs more then "contributor" level for that to work.
Thanks for merging!

Issue #201, opened 2016-08-25
Ship tests in pypi package


Fale:
Please, do ship tests in the pypi artifact

PR #240, opened 2017-01-16
add docs and tests to release tarballs

evgeni:
in the spirit of https://github.com/ansible/ansible/issues/19769 :)


evgeni:
Fixes #201

PR #252, opened 2017-03-01
support listing, retrieving, and updating Configure Tower in Tower setting values

ryanpetrello:


tower_cli/models/base.py
```diff
@@ -733,7 +734,7 @@ def list(self, all_pages=False, **kwargs):
         # Alter the "next" and "previous" to reflect simple integers,
         # rather than URLs, since this endpoint just takes integers.
         for key in ('next', 'previous'):
-            if not response[key]:
+            if not response.get(key):
```
jangsutsr: This alone will not prevent `AttributeError`, `response.get(key,  None)` will.

tower_cli/resources/setting.py
```diff
@@ -0,0 +1,98 @@
...
+        # setting value at a time, so fetch them all and filter
+        try:
+            return next(s for s in self.list()['results'] if s['id'] == pk)
+        except StopIteration:
+            raise exc.NotFound('The requested object could not be found.')
+
+    @resources.command(ignore_defaults=True)
+    def modify(self, pk, value, **kwargs):
+        """Modify an already existing object."""
+        prev_value = self.get(pk)['value']
```
AlanCoding: If this were to behave like other tower-cli modify commands, I think that it would write a debug statement if prev_value==value, and avoid doing the patch.

tower_cli/models/base.py
```diff
@@ -115,10 +115,11 @@ def __new__(cls, name, bases, attrs):
 
         # Ensure that the endpoint ends in a trailing slash, since we
         # expect this when we build URLs based on it.
-        if not newattrs['endpoint'].startswith('/'):
-            newattrs['endpoint'] = '/' + newattrs['endpoint']
-        if not newattrs['endpoint'].endswith('/'):
-            newattrs['endpoint'] += '/'
+        if isinstance(newattrs['endpoint'], basestring):
```
AlanCoding: It looks like some python environment hit `./tower_cli/models/base.py:118:45: F821 undefined name 'basestring'`. I think that jangsutsr was working on a solution to that problem, and he's probably figured it out by now.

tower_cli/resources/setting.py
```diff
@@ -0,0 +1,98 @@
...
+            )
+            e.message = '%s is not a valid category.  Choose from [%s]' % (
+                kwargs['category'],
+                ', '.join(categories)
+            )
+            raise e
+        finally:
+            self.custom_category = None
+        return {
+            'results': [{'id': k, 'value': v} for k, v in result.items()]
```
jangsutsr: A large `value` may cause problems in human-readable output:
```
(venv_for_tower_cli) sitan-OSX:tower-cli sitan$ tower-cli setting get LICENSE
======= ==========================================================================================================================================================================================================================================================================================================================================================================================================
  id                                                                                                                                                                                                      value
======= ==========================================================================================================================================================================================================================================================================================================================================================================================================
LICENSE {u'subscription_name': u'Enterprise Tower up to 100000 Nodes', u'features': {}, u'instance_count': 100000, u'trial': False, u'contact_email': u'gross@weiner.com', u'company_name': u'Dr. James Grossweiner', u'license_type': u'enterprise', u'contact_name': u'Dr. James Grossweiner', u'license_date': 2524608000, u'license_key': u'cf2da1fdbc196c3f9a354d871cdad572dcd2e309a1e49616d8d9c600a9b7fc91'}
======= ==========================================================================================================================================================================================================================================================================================================================================================================================================
```

tower_cli/models/base.py
jangsutsr: That's true. It will fail python3, where `basestring` is removed. This is also the reason for travis test failures. Try using `six.text_type` or whatever is suitable.

tower_cli/models/base.py
ryanpetrello: `response` is a dict, so it would be a `KeyError`.  `dict.get` falls back to `None`, so I think this should work: https://docs.python.org/2/library/stdtypes.html#dict.get

`python -c "print {}.get('missing') is None"`

tower_cli/resources/setting.py
ryanpetrello: I thought about this and one area that this falls apart is encrypted field values - we can't really *tell* if those changed :/

tower_cli/models/base.py
jangsutsr: Indeed. I didn't try `get` without default value before, sorry.

tower_cli/resources/setting.py
AlanCoding: That's true. My personal vote is a no-op in these cases _unless_ it is encrypted (I want to favor fewer database changes with logging settings in mind).

tower_cli/resources/setting.py
```diff
@@ -0,0 +1,98 @@
...
+        """Return one and exactly one object"""
+        # The Tower API doesn't provide a mechanism for retrieving a single
+        # setting value at a time, so fetch them all and filter
+        try:
+            return next(s for s in self.list()['results'] if s['id'] == pk)
+        except StopIteration:
+            raise exc.NotFound('The requested object could not be found.')
+
+    @resources.command(ignore_defaults=True)
+    def modify(self, pk, value, **kwargs):
```
ryanpetrello: Don't we want `use_fields_as_options` so that the value can be specified via `--value` ?

AlanCoding: Here you have `value` as an argument, and if it were an option (of the command like `--value`) I would expected it to be a keyword argument to the function. On the other hand, if you wanted to keep it as an argument you might consider decorating with `@click.argument`.

Also, `use_fields_as_options` can take a list, putting `value` into that would be one possibility.

tower_cli/resources/setting.py
AlanCoding: this also affects just doing `tower-cli setting list`

tower_cli/resources/setting.py
```diff
@@ -0,0 +1,108 @@
...
+
+    @resources.command(ignore_defaults=True)
+    def get(self, pk, **kwargs):
+        """Return one and exactly one object"""
+        # The Tower API doesn't provide a mechanism for retrieving a single
+        # setting value at a time, so fetch them all and filter
+        try:
+            return next(s for s in self.list()['results'] if s['id'] == pk)
+        except StopIteration:
+            raise exc.NotFound('The requested object could not be found.')
```
AlanCoding: I checked out your branch and this works:

`tower-cli setting get SOCIAL_AUTH_SAML_ORGANIZATION_MAP`

This does not work:

`tower-cli setting get --value=SOCIAL_AUTH_SAML_ORGANIZATION_MAP`

That's fine, I somewhat prefer the one that works, but the help text:

```
$ tower-cli setting get
Usage: tower-cli setting get [OPTIONS] [ID]

  Return one and exactly one setting

Options:
  --value TEXT                    [REQUIRED] The value field.
  -h, --tower-host TEXT           The location of the Ansible Tower host.
                                  HTTPS is assumed as the protocol unless
                                  "http://" is explicitly provided. This will
                                  take precedence over a host provided to
                                  `tower config`, if any.
  -u, --tower-username TEXT       Username to use to authenticate to Ansible
                                  Tower. This will take precedence over a
                                  username provided to `tower config`, if any.
  -p, --tower-password TEXT       Password to use to authenticate to Ansible
                                  Tower. This will take precedence over a
                                  password provided to `tower config`, if any.
  -f, --format [human|json|yaml]  Output format. The "human" format is
                                  intended for humans reading output on the
                                  CLI; the "json" and "yaml" formats provide
                                  more data.
  -v, --verbose                   Show information about requests being made.
  --description-on                Show description in human-formatted output.
  --insecure                      Turn off insecure connection warnings. Set
                                  config verify_ssl to make this permanent.
  --certificate TEXT              Path to a custom certificate file that will
                                  be used throughout the command. Overwritten
                                  by --insecure flag if set.
  --help                          Show this message and exit.
```

I have a sense of what your edits to base.py did, and since the other standard use still works I'm fine with that, but you'll see that the help text points us to the `--value` usage. Because of that, use_fields_as_options may need to be revisited.

AlanCoding:
This is giving me trouble now:

```
$ tower-cli setting modify AUTH_LDAP_CONNECTION_OPTIONS '{"OPT_NETWORK_TIMEOUT": 40, "OPT_REFERRALS": 0}'

PATCH https://localhost:8043/api/v1/settings/all/
Params: None
Data: {"AUTH_LDAP_CONNECTION_OPTIONS": "{\"OPT_NETWORK_TIMEOUT\": 40, \"OPT_REFERRALS\": 0}"}

Response: {"AUTH_LDAP_CONNECTION_OPTIONS":["Expected a dictionary of items but got type \"unicode\"."]}

$ tower-cli setting modify AUTH_LDAP_CONNECTION_OPTIONS {"OPT_NETWORK_TIMEOUT": 40, "OPT_REFERRALS": 0}
Usage: tower-cli setting modify [OPTIONS] SETTING [VALUE]

Error: Got unexpected extra arguments (40, OPT_REFERRALS: 0})
```

How can we edit JSON settings?

PR #253, opened 2017-03-02
Add the ability to manage workflows

AlanCoding:
We have basic CRUD options, but workflows nodes have this tricky property that we can only truly address them by their pk value, and nothing else is unique.

To cope with this, I have introduced the workflow `schema` command to obtain both name-based referencing and idempotence. The following is a idempotent schema creation.

```
$ tower-cli workflow schema sch_test @wf_schema.yml
- job_template: 58
  success_nodes:
  - project: 40
```


chrismeyersfsu:
image

Shouldn't the schema be:
```
$ tower-cli workflow schema sch_test @wf_schema.yml
- job_template: 58
  success_nodes:
    - project: 40
```

So that you can support, for example, a tree structure with height > 1.
```
$ tower-cli workflow schema sch_test @wf_schema.yml
- job_template: 58
  success_nodes:
    - project: 40
      success_nodes:
        - job_template: 59
```

AlanCoding:
^ exactly right. Updated code and tests so that it works that way.

This is also important to get a pull-push cycle working. The output should be usable as an input to create the same structure in another workflow.

chrismeyersfsu:
I'd like to see a more complex schema/tree in your single test function.

Maybe 5 nodes, JT1, JT2, P1, I1, 

```
JT1 -> P1 -> JT2 -> JT1
    -> I1
```

AlanCoding:
Very good request. I am still ironing out a few kinks, but after the push I made just now I can demo that.

screen shot 2017-03-02 at 10 45 10 am

```bash
tower-cli workflow schema meyers > meyers_schema.yml
```

This writes the schema to that file, its contents are:

```yaml
- failure_nodes:
  - inventory_source: 42
  job_template: 45
  success_nodes:
  - project: 40
    success_nodes:
    - job_template: 55
      success_nodes:
      - job_template: 44
```

Now I make a new empty workflow:

```bash
tower-cli workflow create --name meyers_copy --description "oh yeah" --organization Default
```

Now, I throw up that entire payload to the new one:

```bash
tower-cli workflow schema meyers_copy @meyers_schema.yml
```

Even better, I run the exact same thing again.

```bash
tower-cli workflow schema meyers_copy @meyers_schema.yml
```

No changes made this round.

chrismeyersfsu:
sexy, 👍

AlanCoding:
Things that have been noticed which are out-of-scope for now, can be future issues:

 - inventory source naming doesn't correspond with group, and we could help the user out by parsing the group name and/or aliasing "group" for "inventory_source"
 - the output isn't colored yellow or green. Doing this will require special handing of the "changed" key, which could also be confusing.

tower_cli/utils/resource_decorators.py
```diff
@@ -0,0 +1,54 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import functools
+from tower_cli.utils import types
+
+import click
+
+
+def unified_job_template_options(method):
```
jangsutsr: Something seems to be not working correctly here. I tested and saw duplicates in help text:
```
...
 -W, --workflow-job-template WORKFLOW
                                  The workflow_job_template field.
  --unified-job-template TEXT     The unified_job_template field.
  -i, --inventory INVENTORY       The inventory field.
  --credential CREDENTIAL         The credential field.
  --job-type TEXT                 The job_type field.
  --job-tags TEXT                 The job_tags field.
  --skip-tags TEXT                The skip_tags field.
  --limit TEXT                    The limit field.
  --inventory-source INVENTORY_SOURCE
                                  Use this inventory source as
                                  unified_job_template field
  --project PROJECT               Use this project as unified_job_template
                                  field
  --job-template JOB_TEMPLATE     Use this job template as
                                  unified_job_template field
  --inventory-source INVENTORY_SOURCE
                                  Use this inventory source as
                                  unified_job_template field
  --project PROJECT               Use this project as unified_job_template
                                  field
  --job-template JOB_TEMPLATE     Use this job template as
                                  unified_job_template field
...
```

tower_cli/resources/node.py
```diff
@@ -0,0 +1,159 @@
...
+        parent_filter_kwargs = {}
+        query_params = ((self._reverse_rel_name(relationship), parent),)
+        parent_filter_kwargs['query'] = query_params
+        if kwargs.get('workflow_job_template', None) is None:
+            parent_data = self.read(pk=parent)['results'][0]
+            parent_filter_kwargs['workflow_job_template'] = parent_data[
+                'workflow_job_template']
+        return parent_filter_kwargs
+
+    @unified_job_template_options
```
jangsutsr: I'm not sure but I have a feeling the cause of duplication lies here.

tower_cli/utils/resource_decorators.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+    Adds the decorators for all types of unified job templates,
+    and if the non-unified type is specified, converts it into the
+    unified_job_template kwarg.
+    """
+    jt_dec = click.option(
```
jangsutsr: Tested to find redundancy w.r.t these three arguments: try creating node with `--job-template $proj_id` creates the node the same way as `--project $proj_id`.

AlanCoding: Yes, that's the point.

tower_cli/resources/node.py
AlanCoding: Good find on the help text.

But the problem isn't here. It was overriding the __getattribute__ that caused this problem... I hoped that would make life simple, but I have dived into more complexity than I bargained for.

tower_cli/resources/node.py
```diff
@@ -0,0 +1,159 @@
...
+}
+
+
+class Resource(models.Resource):
+    cli_help = 'Manage nodes inside of a workflow job template.'
+    endpoint = '/workflow_job_template_nodes/'
+    identity = ('id',)
+
+    workflow_job_template = models.Field(
+        key='-W', type=types.Related('workflow'), required=False)
```
jangsutsr: `workflow_job_template` is a required field in API.

tower_cli/resources/node.py
```diff
@@ -0,0 +1,159 @@
...
+            child_data = self._get_or_create_child(
+                parent, relationship, **kwargs)
+            return child_data
+        return self._assoc(self._forward_rel_name(relationship),
+                           parent, child)
+
+    @resources.command
+    @unified_job_template_options
+    @click.argument('parent', type=types.Related('node'))
+    @click.argument('child', type=types.Related('node'), required=False)
```
jangsutsr: In my opinion, child node should not be omitted in (dis)association (API claims with only one node. Plus, what would be the point just providing one side of a binary relationship?)

AlanCoding: And here I was telling Ryan to use `use_fields_as_options`...

The disassociate pattern should follow the same pattern as all other disassociate commands. Good catch, I need to change a few things here.

tower_cli/resources/node.py
AlanCoding: Carried out this change, and `__new__` looks like it works for this purpose, preventing duplicates.

tower_cli/utils/resource_decorators.py
AlanCoding: Try:

`tower-cli node list --project examples`

and you'll get all the nodes that run the project called "examples". This is still a "lie", but a good one.

PR #257, opened 2017-03-03
Implement timeout field/option for resources that need it

jangsutsr:
Pretty strait-forward implementation, since the API interface was changed just a little.

### DIFF
--- a/tower_cli/resources/inventory_source.py
+++ b/tower_cli/resources/inventory_source.py
@@ -47,6 +47,7 @@ class Resource(models.MonitorableResource):
     # Only used if update_on_launch is used
     update_cache_timeout = models.Field(type=int, required=False,
                                         display=False)
+    timeout =  models.Field(type=int, required=False, display=False)
 
     @click.argument('inventory_source', type=types.Related('inventory_source'))
     @click.option('--monitor', is_flag=True, default=False,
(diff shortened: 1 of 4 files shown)


tower_cli/resources/inventory_source.py
```diff
@@ -47,6 +47,7 @@ class Resource(models.MonitorableResource):
     # Only used if update_on_launch is used
     update_cache_timeout = models.Field(type=int, required=False,
                                         display=False)
+    timeout =  models.Field(type=int, required=False, display=False)
```
AlanCoding: How about some help text telling what the integer should be - seconds vs. minutes

tower_cli/resources/project.py
```diff
@@ -67,6 +69,8 @@ def create(self, organization=None, monitor=False, timeout=None,
         This would be a shared class with user, but it needs the ability
         to monitor if the flag is set.
         """
+        if 'job_timeout' in kwargs and 'timeout' not in kwargs:
```
jangsutsr: Interesting fact here: the `if` clause here is solely to make unit tests happy, because our unit test code does not populate `kwargs` with the complete set of possible items, but in practice it does.

I don't think we need to refactor our unit test code base for this reason alone, but we might need to be careful in mimicking the real world when writing future test codes.

PR #258, opened 2017-03-04
Workflow job summary and monitor

AlanCoding:
This introduces a component of workflow jobs for which I have coined the term "scorecard". This is the workflow analog to standard out. Instead of giving the Ansible standard out stream, we just list the jobs that ran as a part of the workflow. Here is the scorecard in isolation:

```
$ tower-cli workflow_job scorecard 78
Hello World goose    SUCCESSFUL at 2017-03-04T20:04:06. in 15.79      seconds
Hello World goose    SUCCESSFUL at 2017-03-04T20:04:23. in 15.787     seconds
Hello World goose    SUCCESSFUL at 2017-03-04T20:04:46. in 15.401     seconds
Hello World goose    SUCCESSFUL at 2017-03-04T20:05:03. in 15.662     seconds
Hello World goose    SUCCESSFUL at 2017-03-04T20:05:19. in 15.535     seconds

OK. (changed: false)
```

However, I really wanted to leverage this notion in order to get `--monitor` working for workflow jobs. This involved way less work than what you might expect. Here is an example of the output:

screen shot 2017-03-04 at 3 10 24 pm

Each line appears in rolling updates as the jobs run. This does not display a complete representation of the workflow network, these simply come out in chronological order.

This PR has a bunch of extra stuff in it, because this required the work from both the standard out work and the workflow work. The standard out code hasn't been merged yet.

UPDATE: look and feel have changed, see latter comment.

### DIFF
--- a/docs/WORKFLOWS.md
+++ b/docs/WORKFLOWS.md
@@ -175,13 +175,11 @@ sub-branches.
 Use the workflow_job resource to launch workflow jobs.
 This supports the use of extra_vars, which can contain answers to
 survey questions.
-A `--wait` flag is available to poll the server until workflow job
-reaches a completed status. Here is an example of using those features:
+The `--monitor` and `--wait` flag are available to poll the server
+until workflow job reaches a completed status. The `--monitor` option
+will print rolling updates of the jobs that ran as part of the workflow.
+Here is an example of using those features:
 
 ```
-tower-cli workflow_job launch -W "echo Hello World" -e a=1 --wait
+tower-cli workflow_job launch -W "echo Hello World" -e a=1 --monitor
 ```
-
-The `--wait` function will display temporary placeholder text.
-Tower-CLI does not yet offer any streaming updates of events within the
-workflow job.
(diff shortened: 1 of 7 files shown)


AlanCoding:
Need to do:

 - make sure that `--wait` still works
 - update docs

chrismeyersfsu:
`summary`  instead of `scorecard`

AlanCoding:
^^ vote officially recorded.

Looking at the formatting, I'm wondering if we should have some mention of type like JOB, PROJECT UPDATE, INVENTORY UPDATE. I don't see a brief way of fitting this into the columns.

AlanCoding:
Implemented changes based on feedback, new prototype:

screen shot 2017-03-06 at 12 05 31 pm

This took some fighting of the standard formatter, because if we used the code as-is, then it would change column width as new updates rolled in. Thankfully, the unified_job model isn't used for anything else, so I thought it prudent to implement a fixed-width column in that model.

Issue #216, opened 2016-09-30
Support uploading JT survey


AlanCoding:
Although tower-cli supports creating different types of job templates, you can't add a survey. This should be easy to implement as a generic variables field. In addition to this, there is a different endpoint the user needs to hit in order to GET the survey spec. Proposal:
- add `--survey-spec` parameter to job templates
- add `--survey-enabled` boolean parameter, document in survey-spec help text that this needs to be set
- add a `tower-cli job_template get_survey` command to print the survey spec contents

This would be a little clunky to use, but it would get the job done.

AlanCoding:
queued for 3.1.0 release

PR #259, opened 2017-03-07
Add ability to view and add survey specs

AlanCoding:
This is done for JTs here, I need to figure out a way to do them for workflows too. Also did some work making extra_vars processing a little less terrible.

This might be a little early in development to make a PR, but I want to put it out there so people can see the direction I'm going in.

Connect #216

Also thinking might be better to call the GET command "survey"

### DIFF
--- a/tower_cli/models/base.py
+++ b/tower_cli/models/base.py
@@ -36,12 +36,10 @@ from tower_cli import resources
 from tower_cli.api import client
 from tower_cli.conf import settings
 from tower_cli.models.fields import Field
-from tower_cli.utils import exceptions as exc
+from tower_cli.utils import exceptions as exc, parser, debug, secho
 from tower_cli.utils.command import Command
-from tower_cli.utils import debug, secho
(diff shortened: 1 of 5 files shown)


tower_cli/models/base.py
```diff
@@ -1208,3 +1206,45 @@ def modify(self, pk=None, create_on_missing=False, **kwargs):
...
+                r = client.post(self._survey_endpoint(ret['id']),
+                                data=survey_input)
+            if r.status_code == 200:
+                ret['changed'] = True
+            if survey_input and not ret['survey_enabled']:
+                debug.log('For survey to take effect, set survey_enabled'
+                          ' field to True.', header='warning')
+        return ret
+
+    @resources.command
```
jangsutsr: One of the values of model/command-line separation comes here: `pk` is the only field needed for `survey` command, but all model fields automatically fill `kwargs`, which makes code confusing and pollute help text.

Issue #260, opened 2017-03-07
inventory_script create does not accept non-ASCII characters in script


ldelouw:
[root@tower ~]# file inventory.py 
inventory.py: Python script, UTF-8 Unicode text executable

[root@tower ~]# tower-cli inventory_script create  -n generic-dummy --organization org-generic --script "$(cat ~/inventory.py)"

Traceback (most recent call last):
  File "/usr/bin/tower-cli", line 90, in <module>
    cli()
  File "/usr/lib/python2.7/site-packages/click/core.py", line 716, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File "/usr/bin/tower-cli", line 82, in invoke
    return super(TowerCLI, self).invoke(ctx)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python2.7/site-packages/click/core.py", line 1058, in invoke
    sub_ctx = cmd.make_context(cmd_name, args, parent=ctx)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 620, in make_context
    self.parse_args(ctx, args)
  File "/usr/lib/python2.7/site-packages/tower_cli/utils/command.py", line 43, in parse_args
    return super(Command, self).parse_args(ctx, args)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 874, in parse_args
    value, args = param.handle_parse_result(ctx, opts, args)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 1390, in handle_parse_result
    value = self.full_process_value(ctx, value)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 1675, in full_process_value
    return Parameter.full_process_value(self, ctx, value)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 1359, in full_process_value
    value = self.process_value(ctx, value)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 1349, in process_value
    return self.type_cast_value(ctx, value)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 1338, in type_cast_value
    return _convert(value, (self.nargs != 1) + bool(self.multiple))
  File "/usr/lib/python2.7/site-packages/click/core.py", line 1336, in _convert
    return self.type(value, self, ctx)
[...]

PR #261, opened 2017-03-07
Fix issue260

jangsutsr:
Connect #260.

Note here the encoding is tacitly assumed to be utf-8. It is safe in most situations though,  considering the prevalence of utf-8 encoding nowadays.


AlanCoding:
Ouch, I was worried that the strange behavior was because it needed to be decoded vs encoded. Once we get this tested it should be g2g.

ldelouw:
Tested, works well :+1:

Issue #256, opened 2017-03-02
tower-cli schedule create does not accept PATCH HTTP method


bp8017:
Tower CLI 3.0.2

```
tower-cli schedule create --name "Example Schedule" --description "Example Schedule" --job-template 288 --rrule "DTSTART:20170302T110000Z RRULE:FREQ=HOURLY;INTERVAL=1" -u admin -p 'XXXXXXX' --force-on-exists

Error: The Tower server says you can't make a request with the PATCH method to that URL (https://127.0.0.1/api/v1/job_templates/288/schedules/7/).
```

AlanCoding:
`/api/v1/job_templates/288/schedules/7/` looks like an invalid URL.

bp8017:
It looks like the extra "--force-on-exist" does not work with the schedule create subcommand. However, it would be nice if it allowed for some level of idempotency by running it again with the same arguments as the other subcommands support.

jangsutsr:
@bp8017 Thx for the bug report. It is fixed and will be shipped with 3.1.

PR #263, opened 2017-03-07
Resolve issue256

jangsutsr:
Connect #256.

This issue reveals a corner case when creating a dependent resource with `--force-on-exists` on:

Suppose resource B can only be created on top of resource A (`schedule` on top of `unified job templates` in this case), The endpoint for creation is thus `/A/pkA/B/`. The modification endpoint, on the other hand, is `/B/pkB/`. A `create` method with `--force-on-exists` on converts creation to modification if the resource B exists, and the PATCH endpoints is simply concatenated into `/A/pkA/B/pkB/`, whose format is not a valid API *for now*.

The solution is dependent on API convention. Considering the possibility of further API convention changes (although highly unlikely ---- it makes little sense to have endpoints for individual related resource), the implementation here might change accordingly.

### DIFF
--- a/tower_cli/models/base.py
+++ b/tower_cli/models/base.py
@@ -544,6 +544,11 @@ class ResourceMethods(BaseResource):
         # Return the response.
         return resp
 
+    def _get_patch_url(self, url, pk):
+        """Overwrite this method to handle specific corner cases to
+        the url passed to PATCH method."""
+        return url + '%s/' % pk
+
     def write(self, pk=None, create_on_missing=False, fail_on_found=False,
               force_on_exists=True, **kwargs):
         """Modify the given object using the Ansible Tower API.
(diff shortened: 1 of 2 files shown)


AlanCoding:
I kind of don't like this. It's just hard to read and if I happened to come by this point in code I wouldn't really be able to guess what it was doing.

I would prefer to outright refactor this:

```
        # Get the URL and method to use for the write.
        url = self.endpoint
        method = 'POST'
        if pk:
            url += '%s/' % pk
            method = 'PATCH'

        # If debugging is on, print the URL and data being sent.
        debug.log('Writing the record.', header='details')

        # Actually perform the write.
        r = getattr(client, method.lower())(url, data=kwargs)
```

Even this isn't super clear to read. What if we introduced something like a `get_xxxx_url` method, put in in the ResourceMethods class, and then over-ride this in the schedule class?

Refactor example:

```
        # Get the URL and method to use for the write.
        url = self.get_post_url(pk)
        method = 'POST'
        if pk:
            url += self.get_patch_url(pk)
            method = 'PATCH'

        # If debugging is on, print the URL and data being sent.
        debug.log('Writing the record.', header='details')

        # Actually perform the write.
        r = getattr(client, method.lower())(url, data=kwargs)
```

PR #264, opened 2017-03-08
Make help text for --extra-vars better. it can be used multiple time …

ldelouw:
…to add multiple variables. Useful when scripting

Simple change.

Thanks


AlanCoding:
Thanks for the contribution. I merged this into #259 to resolve conflicts, and we should get that merged into master soon.

PR #265, opened 2017-03-08
Expose --wait option for projects and groups

AlanCoding:
In the process of testing out the master branch, I noticed this. This makes it so that commands like this work:

`tower-cli project create --name="dup2" --organization "Default" --scm-type git --scm-url https://github.com/AlanCoding/permission-testing-playbooks.git --wait`

We had it for jobs, and most of the code was already in place but they also needed to be added as options.


AlanCoding:
My test fix was:

```python
self.assertIn((42,), isrc_sync.call_args)
```

I'd like to use this pattern in more places to avoid more duplicate code, because we cary through a lot of redundant kwargs that aren't relevant.

PR #266, opened 2017-03-09
Change the ordering in which runtime config vars are processed

AlanCoding:
This is an extremely aggressive change, which gets rid of an old multi-purpose decorator (which was confusingly named "command", like many other command decorators), and moves one of its functions (the application of runtime parameters) into a callback method. Also relies on `is_eager`.

One shortcoming might be the removal of the contextmanager, which could mean that the values are "sticky" from one operation to another if used in a python context. I think the right solution going from there would be to just do some post-action hook of nuking the runtime config parser.

This should resolve a lot of reported issues simultaneously, so I'm not going to go about listing them all just yet. But there are some serious release planning considerations, and I want to give @wenottingham an idea that, no, this is not at all a simple fix, and merging this will cary substantial risks of unforeseen bugs, although the bug this resolves is likewise rather severe.

We probably want to exclude this for 3.1.0. I won't be able to do the necessary followup work (you can bet some tests will fail) until after next week. I'd vote to ship this this in a quick 3.1.1.

For the record, this problem also looks like it has existed for the vast majority of the history of tower-cli.

### DIFF
--- a/tests/test_utils_decorators.py
+++ b/tests/test_utils_decorators.py
@@ -15,35 +15,33 @@
 
 import click
 
-from tower_cli.utils.decorators import command
-
 from tests.compat import unittest
+
+from tower_cli.utils import decorators
 
 
 class CommandTests(unittest.TestCase):
     """Establish that the @command decorator works as I expect,
     with and without a call signature.
     """
-    def test_command_without_call_signature(self):
-        """Establish that if we call the @command decorator without a
-        call signature, that it decorates the function appropriately.
-        """
-        # Define a command.
-        @command
-        def foo():
-            pass
-
-        # Ensure that it's a command.
-        self.assertIsInstance(foo, click.core.Command)
-
     def test_command_with_call_signature(self):
-        """Establish that if we call the @command decorator with a call
+        """Establish that if we call the @click.command decorator with a call
         signature, that it decorates the function appropriately.
         """
         # Define a command
-        @command()
+        @click.command()
         def foo():
             pass
 
         # Ensure that it's a command.
         self.assertIsInstance(foo, click.core.Command)
+
+    def test_runtime_context_manager(self):
+        """Test that the kwargs from settings are removed before
+        running the command from the resource itself.
+        """
+        def foo(**kwargs):
+            self.assertEqual(kwargs, {})
+
+        foo = decorators.runtime_context_manager(foo)
+        foo(tower_username='foobear')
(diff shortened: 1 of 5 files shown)


AlanCoding:
Some more considerations are coming to light...

I added in some code to no longer make them "sticky", such that any existing runtime parameters are removed when invoking a command.

I have noticed that with this change, the settings get passed into `kwargs` for any arbitrary command. This will probably have undesirable effects, so this is a big TBD. Unfortunately, I can't think of a good way to resolve this except for re-introducing another decorator like the one I just removed. However, the contextmanager won't work for the original stated reasons here. The thing we really need is a post-command hook to destroy the runtime settings. Previously, the context manager would save the old runtime settings and reapply them after the command. To me, this is a bridge too far because what are "runtime" parameters other than parameters that don't persist?

The resolution order of decorators are confusing here, because click manages the order of processing of its own options, however, it still seems that a custom decorator can still be implemented and will be next-in-line _after_ click option processing. This was implicit in the previous runtime settings context manager, although you couldn't call it logically consistent considering that some options should have required the settings to already be set.

We also have this warning:

```
/Users/alancoding/.virtualenvs/cli2/lib/python2.7/site-packages/ansible_tower_cli-3.1.0-py2.7.egg/tower_cli/models/base.py:256: Warning: Click detected the use of the unicode_literals __future__ import.  This is heavily discouraged because it can introduce subtle bugs in your code.  You should instead use explicit u"" literals for your unicode strings.  For more information see http://click.pocoo.org/python3/
  new_method)
```

http://click.pocoo.org/5/python3/

Maybe we can simply apply the suggestion like `io.BytesIO(input.encode('utf-8'))` to something, but I'm still not sure why we see this now when we didn't before.

AlanCoding:
Ok, pushed a fix for the biggest of the issues I mentioned. The settings options are no longer passed into the methods themselves, and I destroy the runtime settings entirely at the end of a command.

AlanCoding:
Well, I just outright removed the import of `unicode_literals` from base.py, and everything seems to still be working correctly, even giving unicode fields as options.

I'm pretty confident in saying that I know why we hit the warning now, but not earlier. The warning is triggered if we import that _inside of the file that applies the command decorator from click_. In my changes, I moved the application of the decorator from the utils to base.py. The utils file did not have the import, thus the change. It was presumably a bad practice in all locations where it was applied, which includes a vast array of files throughout the code base.

Issue #269, opened 2017-03-13
unicode brokenness in 3.1.0


wenottingham:
```
[notting@nostromo: ~]$ tower-cli job_template list
== ============================ ========= ======= ======================= 
id             name             inventory project        playbook         
== ============================ ========= ======= ======================= 
15 Apply standard configuration         5       7 apply-configuration.yml
12 Deploy application                   5       7 deploy.yml
14 Destroy cloud instances              5       7 demo-aws-delete.yml
17 fred                                 5       7 demo-aws-delete.yml
13 Provision cloud instances            5       7 demo-aws-launch.yml
== ============================ ========= ======= ======================= 
[notting@nostromo: ~]$ tower-cli job_template get 17
== ==== ========= ======= =================== 
id name inventory project      playbook       
== ==== ========= ======= =================== 
17 fred         5       7 demo-aws-delete.yml
== ==== ========= ======= =================== 
[notting@nostromo: ~]$ tower-cli job_template delete 17
Traceback (most recent call last):
  File "/usr/bin/tower-cli", line 90, in <module>
    cli()
  File "/usr/lib/python2.7/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/bin/tower-cli", line 82, in invoke
    return super(TowerCLI, self).invoke(ctx)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python2.7/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/tower_cli/utils/decorators.py", line 51, in answer
    return method(*inner_a, **inner_kw)
  File "/usr/lib/python2.7/site-packages/tower_cli/models/base.py", line 309, in func
    result = method(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/tower_cli/models/base.py", line 695, in delete
[...]

PR #270, opened 2017-03-13
Use safer format indicator across base.py

jangsutsr:
connect #269.

It's terrifying to think such a bug has been lurking in base.py for so long: [Our `pk` argument is initialized as a string](https://github.com/ansible/tower-cli/blob/master/tower_cli/models/base.py#L263), yet `%d` is used to represent `pk` across various places in base.py. It is okay if `pk` is substituted by a real integer beforehand, but will cause traceback if it keeps what it is.

A safer way is substituting `%s` over `%d`, which accepts integer types as well. Guess the initial purpose of using `%d` is for type check, but a server-side error is way better than a traceback *_*

### DIFF
--- a/tower_cli/models/base.py
+++ b/tower_cli/models/base.py
@@ -692,7 +692,7 @@ class ResourceMethods(BaseResource):
         # If it turns out the record doesn't exist, handle the 404
         # appropriately (this is an okay response if `fail_on_missing` is
         # False).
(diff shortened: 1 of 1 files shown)


AlanCoding:
Nice job fixing all those occurrences!

If we ever create a `CONTRIBUTING.md` guide, then I could see using a section about unicode with notes about every case like this we encounter.

Issue #276, opened 2017-03-21
--extra-vars not working anymore


sebastiangraf:
In tower-cli 3.1.0 and 3.2.0, creating job_templates with --extra-vars is not working any more:

`tower-cli job_template create --name was85-nets_cote_nightly-dev-netsc-startcluster --job-type run --inventory was85-dev-netsc --project Plattformen-SBB --playbook was85_default_start.yml --machine-credential was85-dev --ask-variables-on-launch False --extra-vars project=nets_cote_nightly --force-on-exists
`
results in a Job Template where the extra-vars is empty. I tried with `--extra-vars {project:nets_cote_nightly} ` which was not working either.

AlanCoding:
Fallout from #216, and collision between the `create` overwrite for the `job_type` parameter with the overwrite of `write` for the `extra_vars` parameter. Although, the exact mechanism by which this collision happened is still unclear to me, it has always been difficult to predict the behavior when using patterns like that.

AlanCoding:
planning to ship fix in 3.1.3 relatively soon.

PR #277, opened 2017-03-21
Fix bug where extra_vars were lost on create

AlanCoding:
Related to #276

This does a few things, I need to come back and write up some details later.

### DIFF
--- a/tower_cli/models/base.py
+++ b/tower_cli/models/base.py
@@ -484,7 +484,8 @@ class ResourceMethods(BaseResource):
         click is unfortunately bad at the way it sends through unspecified
         defaults."""
         for key, value in copy(kwargs).items():
-            if value is None:
+            # options with multiple=True return a tuple
+            if value is None or value == ():
                 kwargs.pop(key)
             if hasattr(value, 'read'):
                 kwargs[key] = value.read()
@@ -1204,18 +1205,14 @@ class Resource(ResourceMethods):
                   help='If used, if a match is found on unique fields, other '
                        'fields will be updated to the provided values. If '
                        'False, a match causes the request to be a no-op.')
-    def create(self, fail_on_found=False, force_on_exists=False, **kwargs):
+    def create(self, **kwargs):
         """Create an object.
 
         Fields in the resource's `identity` tuple are used for a lookup;
         if a match is found, then no-op (unless `force_on_exists` is set) but
         do not fail (unless `fail_on_found` is set).
         """
-        return super(Resource, self).write(
-            create_on_missing=True,
-            fail_on_found=fail_on_found, force_on_exists=force_on_exists,
-            **kwargs
-        )
+        return self.write(create_on_missing=True, **kwargs)
 
     @resources.command(ignore_defaults=True)
     @click.option('--create-on-missing', default=False,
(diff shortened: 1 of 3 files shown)


tower_cli/models/base.py
```diff
@@ -484,7 +484,8 @@ def _pop_none(self, kwargs):
         click is unfortunately bad at the way it sends through unspecified
         defaults."""
         for key, value in copy(kwargs).items():
-            if value is None:
+            # options with multiple=True return a tuple
+            if value is None or value == ():
```
AlanCoding: I said I would prove more detail, so I'll do it in-line:

This is needed to more fully support the use of multiple=True in the case of extra_vars or other fields. Options that allow multiples always return as a tuple. If an input value is a tuple type and is empty, then we can be confident that click is telling us "the user has not specified any values".

tower_cli/resources/job_template.py
```diff
@@ -99,17 +99,12 @@ class Resource(models.SurveyResource):
         help_text='On write commands, perform extra POST to the '
                   'survey_spec endpoint.')
 
-    @resources.command
-    def create(self, fail_on_found=False, force_on_exists=False,
-               extra_vars=None, **kwargs):
-        """Create a job template."""
+    def write(self, *args, **kwargs):
```
AlanCoding: Here, `create` is being converted into `write`. The user of the former was what caused the bug in the first place. The standard options were getting applied, but somehow were not functioning correctly. Over-riding the write method is extremely reliable by comparison since it is strictly used internally (not as a command), so its behavior will be predictable as the standard python behavior even if each level is using stuff from its super-class.

tower_cli/resources/job_template.py
```diff
@@ -99,17 +99,12 @@ class Resource(models.SurveyResource):
...
         # Provide a default value for job_type, but only in creation of JT
-        if not kwargs.get('job_type', False):
+        if (kwargs.get('create_on_missing', False) and
```
AlanCoding: We force `job_type` to be "run", because tower-cli actually doesn't have a good system of defaults for fields. If we just set `default="run"` for the job_type field, then on any modify command it will _change_ the job_type to run, which is extremely undesirable. We want to interpret this kind of default as only applying when creating a new resource. This can be done with by the create command, or by the write command with --create-on-missing set.

tower_cli/models/base.py
```diff
@@ -1204,18 +1205,14 @@ class Resource(ResourceMethods):
...
         Fields in the resource's `identity` tuple are used for a lookup;
         if a match is found, then no-op (unless `force_on_exists` is set) but
         do not fail (unless `fail_on_found` is set).
         """
-        return super(Resource, self).write(
-            create_on_missing=True,
-            fail_on_found=fail_on_found, force_on_exists=force_on_exists,
-            **kwargs
-        )
+        return self.write(create_on_missing=True, **kwargs)
```
AlanCoding: This is just consolidating code here. Even if you give explicit default, click will still force `fail_on_found=None` if the user didn't provide a value, so we're just repeating ourselves.

tower_cli/resources/job_template.py
jangsutsr: ```The standard options were getting applied, but somehow were not functioning correctly.```

I think I figured it out: It's all in the extra `extra_vars=None`, if you try doing nothing, but simply deleting that part, you will find the issue solved too.

PR #278, opened 2017-03-22
implement relaunch of jobs like in the UI

evgeni:


tower_cli/models/base.py
```diff
@@ -1189,6 +1189,30 @@ def cancel(self, pk=None, fail_if_not_running=False, **kwargs):
...
+        Fails with a non-zero exit status if the job cannot be relaunched.
+        You must provide either a pk or parameters in the job's identity.
+        """
+        # Search for the record if pk not given
+        if not pk:
+            existing_data = self.get(**kwargs)
+            pk = existing_data['id']
+
+        relaunch_endpoint = '%s%s/relaunch/' % (self.endpoint, pk)
+        data = {'extra_vars': {}}
```
AlanCoding: seems like this could just be an empty dict

tower_cli/models/base.py
```diff
@@ -1189,6 +1189,30 @@ def cancel(self, pk=None, fail_if_not_running=False, **kwargs):
...
+        # Attempt to relaunch the job.
+        try:
+            client.post(relaunch_endpoint, data=data)
```
AlanCoding: It would be great if we can store this to a variable, get the `.json()` version and then put into the dict the method returns. Old versions may not support this, so maybe add a simple check that "id" is in the response data before using it.

tower_cli/models/base.py
```diff
@@ -1189,6 +1189,30 @@ def cancel(self, pk=None, fail_if_not_running=False, **kwargs):
...
+            changed = True
+        except exc.MethodNotAllowed:
+            changed = False
+
+        # Return a success.
+        return {'status': 'relaunched', 'changed': changed}
```
AlanCoding: If you do the ^ stuff, then best to remove "status" from this, so we use the API's value.

evgeni:
:)

Last update adds a small test too :)

PR #280, opened 2017-03-23
implement copying of resources as in the UI

evgeni:


tower_cli/models/base.py
```diff
@@ -1215,6 +1215,22 @@ def create(self, **kwargs):
...
+        orig = self.read(pk, fail_on_no_results=True,
+                         fail_on_multiple_results=True)
+        orig = orig['results'][0]
+        # Remove default values (anything where the value is None).
+        self._pop_none(kwargs)
+
+        newresource = copy(orig)
+        newresource.update(kwargs)
+        newresource['id'] = None
+        newresource['name'] += " @ %s" % (time.strftime('%X'))
```
AlanCoding: Slight nuanced case - making a copy of a copy. Try using this instead:

```python
newresource['name'] = newresource['name'].split('@', 1)[0] + ' @ ' + time.strftime('%I:%M:%S %p')
```

Let me know if that doesn't work for you. Stylistically you might want to alter it to use the `%` synatx.

tower_cli/models/base.py
```diff
@@ -1215,6 +1215,22 @@ def create(self, **kwargs):
...
+        """
+        orig = self.read(pk, fail_on_no_results=True,
+                         fail_on_multiple_results=True)
+        orig = orig['results'][0]
+        # Remove default values (anything where the value is None).
+        self._pop_none(kwargs)
+
+        newresource = copy(orig)
+        newresource.update(kwargs)
+        newresource['id'] = None
```
AlanCoding: Maybe just `.pop('id')` instead?

tower_cli/models/base.py
```diff
@@ -1215,6 +1215,22 @@ def create(self, **kwargs):
...
+        """Copy an object.
+        """
+        orig = self.read(pk, fail_on_no_results=True,
+                         fail_on_multiple_results=True)
+        orig = orig['results'][0]
+        # Remove default values (anything where the value is None).
+        self._pop_none(kwargs)
+
+        newresource = copy(orig)
+        newresource.update(kwargs)
```
AlanCoding: This over-ride functionality is meaningful for properties not in the identity tuple, I think that's fine but not sure how to document. I'd mention in the command's help text that it applies user-provided params to the _new_ object.

tower_cli/models/base.py
evgeni: ack

tower_cli/models/base.py
evgeni: done

tower_cli/models/base.py
evgeni: yeah, good catch.

any reason you used explicit `time.strftime('%I:%M:%S %p')` and not `%X` which should (at least sometimes) match the user locale?

AlanCoding: The main reason for that was to have the am/pm for purely aesthetic reasons.

AlanCoding: If `%X` is more internationally compliant, I would say leave it as is.

PR #282, opened 2017-03-26
Fix the source name in the RPM SPEC

ssato:
Fixes for the RPM packaging stuff:

- fix the source name in the RPM SPEC to make it build-able.
- make tests/**/*.py not incuded as a python package; it feels unclear but directly related to the RPM also, that is, %python_sitelib/tests/ will be made by this RPM if it's not.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,6 @@
 import re
 import sys
 from distutils.core import setup
-from setuptools import find_packages
 from setuptools.command.test import test as TestCommand
 
 
@@ -131,8 +130,9 @@ setup(
     scripts=[
         'bin/tower-cli',
     ],
-    packages=find_packages(),
-
+    packages=[
+        'tower_cli',
+    ],
     # How to do the tests
     tests_require=['tox'],
     cmdclass={'test': Tox},
--- a/packaging/rpm/ansible-tower-cli.spec
+++ b/packaging/rpm/ansible-tower-cli.spec
@@ -1,13 +1,14 @@
 %{!?python_sitelib: %define python_sitelib %(%{__python} -c "from distutils.sysconfig import get_python_lib; print get_python_lib()")}
+%global srcname tower-cli
 
-Name:           ansible-tower-cli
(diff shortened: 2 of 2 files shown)


AlanCoding:
Thanks for the contribution!

> make tests/**/*.py not incuded as a python package;

I'm not automatically following where this happens in your PR. I am reminded of #240.

I would like @wenottingham to have a look-over, and also get @ryanpetrello appraisal of the change in the use of `find_packages`.

wenottingham:
The spec stuff looks OK as long as we're naming the RPM ansible-tower-cli while just releasing a tarball called tower-cli. I could see making them match, but the mostly likely way to do tht is renaming the RPM package.

setup.py
```diff
@@ -131,8 +130,9 @@ def combine_files(*args):
     scripts=[
         'bin/tower-cli',
     ],
-    packages=find_packages(),
-
+    packages=[
+        'tower_cli',
+    ],
```
AlanCoding: I tried testing this by doing `sudo python setup.py sdist`, and I found that both docs and tests were still getting included.

Sidebar: if we do exclude tests, I would prefer to use the syntax of `find_packages(exclude=['tests'])`

If I remove tests from `MANIFEST.in`, then I find that it really does not include the folder.

However, I don't feel like we have universal agreement that the tests should not be included, see #201. So I think the best course of action for the moment would be to role back the change this this line. As of yet I don't think it's having an effect anyway.

ssato: Thanks for the review!

The problem is not that tests/ included in src distribution but bin distribution. For example, the result rpm by 'python setup.py bdist_rpm' (or archive by '... bdist') includes '.../site-packages/tests/...'. This is the problem.

ssato: And of course 'find_packages(exclude=['tests'])' may work but I just thought and felt that it might be a little excessive as only tower_cli to install as a module.

ssato:
Thanks a lot for quick response!

PR #285, opened 2017-03-27
For compatibility, also convert null-ish data into empty dict

AlanCoding:
This will give enhanced compatibility across various Tower versions. Otherwise, actions like project update, launches, etc. could selectively break in some systems at some times.


AlanCoding:
Doesn't appear that this is needed.

Issue #273, opened 2017-03-15
Allow auth-token Authentication for Tower-CLI Request


jamesmarshall24:
Request to implement the ability to request an auth token once, cache locally, and check expiration on each run. At this time, each tower-cli command ran creates a new auth token. Implementing auth tokens would decrease the number of authentication requests when heavily utilizing tower-cli.

PR #303, opened 2017-06-05
Implement token-based authentication

jangsutsr:
Connect #273.


tower_cli/api.py
```diff
@@ -18,16 +18,69 @@
...
+                'headers': {
+                    'Content-Type': 'application/json'
+                }
+            }
+        ).json()
+
+    def _get_auth_token(self):
+        try:
+            with open('/tmp/tower_cli_token.json') as f:
+                token_json = json.load(f)
```
AlanCoding: I think one requirement for this will be that this file isn't world-readable. For that, I think it would be good to have other eyes on it aside from mine.

Aside from that detail, I'm very happy with the your implementation here, managing the expiration and falling back to create a new one.

tower_cli/api.py
```diff
@@ -18,16 +18,69 @@
...
+            if (not isinstance(token_json, dict) or
+                'token' not in token_json or 'expires' not in token_json or
+                dt.utcnow() > dt.strptime(token_json['expires'],
+                                          TOWER_DATETIME_FMT)):
+                raise
+            return 'Token ' + token_json['token']
+        except Exception:
```
AlanCoding: Nonetheless, a predictable request would be to list the exceptions. This could be IOError + a custom exception in the event that the token is expired.

tower_cli/api.py
AlanCoding: One suggestion for the location of this file might be `/var/lib/tower_cli_token.json`?

https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard

> State information. Persistent data modified by programs as they run, e.g., databases, packaging system metadata, etc.

shanemcd: The user running tower-cli would need to have write access to `/var`, which they probably won't unless they're running as root. Which they probably aren't.

wenottingham: Still, don't use a predictable one-off filename in /tmp. Put it in $HOME, and/or use mkstemp()

shanemcd: Maybe consider `/var/tmp/`

shanemcd: @jangsutsr @AlanCoding Here is the Python wrapper for what @wenottingham mentioned: https://docs.python.org/2/library/tempfile.html

AlanCoding: I'm not sure how mkstemp() could work for this case, because it would just shift the problem to having to recall the folder name.

jangsutsr: @AlanCoding That's right, and the only place for memorization is `tower-cli.cfg`, which we might probably not want to contaminate with generated data. `$HOME/.tower_cli_token.json` should be the optimal solution for now.

PR #307, opened 2017-06-08
Refactor to have cli directory.

jangsutsr:
First step towards transferring Tower CLI from CLI to CLI + Python lib.


AlanCoding:
In the release notes of the version this goes out in, I want to specifically call out the deprecation of `as_command`, and note the new class for subcommands. Also, specifically note that `get_resource` will continue to work as usual (this should be the most common use pattern).

Those have some risk of impacting people who have built off of tower-cli, but I think it's sufficiently small, and I'm good with shipping this.

Issue #249, opened 2017-02-28
Provisioning Callbacks


wtcross:
As an Ansible Tower user I should be able to utilize the Tower CLI for triggering provisioning callbacks, rather than relying on a script like those distributed with Tower (`request_tower_configuration.{sh,ps1}`). The provisioning callback feature of the Tower CLI should support not only triggering the callback, but also have proper retry logic and other features of the distributed scripts.

AlanCoding:
This will probably be a new `callback` command in the job_template.py file.

I'm not sure exactly what errors we should retry on, because 400 errors don't offer any hope to success on subsequent retries, but we might just catch any error for simplicity, or emulate the example script as-is.

PR #308, opened 2017-06-20
Implement provisioning callback

jangsutsr:
Connect #249.

### DIFF
--- a/tower_cli/resources/job_template.py
+++ b/tower_cli/resources/job_template.py
@@ -18,7 +18,8 @@ from __future__ import absolute_import, unicode_literals
 import click
 
 from tower_cli import models, resources
-from tower_cli.utils import types
+from tower_cli.utils import types, parser
+from tower_cli.api import client
 
 
 class Resource(models.SurveyResource):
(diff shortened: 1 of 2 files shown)


tower_cli/resources/job_template.py
```diff
@@ -145,3 +146,18 @@ def disassociate_notification_template(self, job_template,
...
                               job_template, notification_template)
+
+    @resources.command(use_fields_as_options=('extra_vars'))
+    @click.option('--host-config-key', help='Job-template-specific string used to authenticate '
+                  'host during provisioning callback.')
+    def callback(self, pk=None, host_config_key='', extra_vars=None):
+        """Contact Tower and request a configuration update using this job template."""
+        url = self.endpoint + '%s/callback/' % pk
+        if not host_config_key:
+            host_config_key = client.get(url).json()['host_config_key']
```
AlanCoding: so the big question here is if this authentication _by itself_ will be sufficient to authenticate the request, even if the tower-cli config is wrong. Docs:

http://docs.ansible.com/ansible-tower/2.2.0/html/userguide/job_templates.html#provisioning-callbacks

```bash
root@localhost:~$ curl --data "host_config_key=5a8ec154832b780b9bdef1061764ae5a" \
                    http:///api/v1/job_templates/1/callback/
```

This seems to suggest that it would. Even if (wrong) basic auth parameters are given along with it, I would put my money on it working.

jangsutsr: It is true that provisioning callback does not need require normal Tower authentication at all. The latest patch solves the issue by explicitly set `auth=None`.

PR #309, opened 2017-06-23
add stable API version detection

ryanpetrello:


tests/test_api.py
```diff
@@ -41,14 +41,18 @@ def test_prefix_implicit_https(self):
         URL prefix given a host with no specified protocol.
         """
         with settings.runtime_values(host='33.33.33.33'):
-            self.assertEqual(client.prefix, 'https://33.33.33.33/api/v1/')
+            with mock.patch.object(Client, '_make_request') as req:
```
ryanpetrello: Now prefix detection itself makes an HTTP request, so I had to mock it to return the expected JSON content from `/api/`.

tests/test_api.py
```diff
@@ -41,14 +41,18 @@ def test_prefix_implicit_https(self):
...
+                req.return_value.json.return_value = {'current_version': '/api/v1/'}
```
ryanpetrello: Same here.

tests/test_api.py
```diff
@@ -41,14 +41,18 @@ def test_prefix_implicit_https(self):
...
+                self.assertEqual(client.prefix, 'https://33.33.33.33/api/v1/')
 
     def test_prefix_explicit_protocol(self):
         """Establish that the prefix property returns the appropriate
         URL prefix and don't clobber over an explicit protocol.
         """
         with settings.runtime_values(host='bogus://33.33.33.33/'):
-            self.assertEqual(client.prefix, 'bogus://33.33.33.33/api/v1/')
+            with mock.patch.object(Client, '_make_request') as req:
```
ryanpetrello: Same here.

tests/test_api.py
```diff
@@ -106,10 +110,9 @@ def test_connection_ssl_error_verbose(self):
             with settings.runtime_values(verbose=True, host='https://foo.co'):
                 with mock.patch.object(Session, 'request') as req:
                     req.side_effect = ErrorType
-                    with mock.patch.object(debug, 'log') as dlog:
+                    with mock.patch.object(debug, 'log') as dlog:  # noqa
                         with self.assertRaises(exc.ConnectionError):
                             client.get('/ping/')
-                        self.assertEqual(dlog.call_count, 5)
```
ryanpetrello: Because there's an additional request (to determine the API version), the number of log lines changed.  But honestly, checking the number of SSL-related log lines is kind of a silly and brittle thing to test.

tests/test_api.py
```diff
@@ -106,10 +110,9 @@ def test_connection_ssl_error_verbose(self):
             with settings.runtime_values(verbose=True, host='https://foo.co'):
                 with mock.patch.object(Session, 'request') as req:
                     req.side_effect = ErrorType
-                    with mock.patch.object(debug, 'log') as dlog:
+                    with mock.patch.object(debug, 'log') as dlog:  # noqa
```
ryanpetrello: As best as I can tell, this patch is really only here to keep the tests from spewing out a bunch of SSL warnings in the test output.

tests/test_api.py
```diff
@@ -170,17 +173,12 @@ def test_insecure_connection(self):
...
         with mock.patch('requests.sessions.Session.request') as g:
-            mock_response = type('statobj', (), {})()  # placeholder object
-            mock_response.status_code = 200
-            g.return_value = mock_response
-            with client.test_mode as t:
-                t.register('/ping/', "I'm a teapot!", status_code=200)
+            with client.test_mode:
                 with settings.runtime_values(verify_ssl=False):
-                    client.get('/ping/')
+                    client._make_request('GET', '/api/', [], {})
```
ryanpetrello: This patching was pretty complicated beforehand - it was relying on some very specific implementation details of the requests library to simulate fake response objects.  It was also hard to mock properly because now there's two requests being made (one to determine the API version, and the other for `/api/v1/ping/`.

So I've simplified this test to really just perform the API version detection request so that we can make sure that `verify=False` was used properly (which is the whole point of this test).

tower_cli/api.py
```diff
@@ -262,9 +266,27 @@ def test_mode(self):
...
             adapters = copy.copy(self.adapters)
-            faux_adapter = FauxAdapter(
-                url_pattern=self.prefix.rstrip('/') + '%s',
-            )
+
+            class VersionedFauxAdapter(FauxAdapter):
+                # An adapter that doesn't track version detection requests
+                # (those sent to /api/
+                def send(self, request, stream=False, **kwargs):
+                    ret = super(VersionedFauxAdapter, self).send(request, stream, **kwargs)
```
ryanpetrello: the FauxAdapter tracks a list of requests made to it, and we have *hundreds* of tests that make assertions about the tests contents - and reference them by index.  So we'll have tests that do things like:

The *first* request is to `/api/v2/xyz/`, the *second* request is a `POST` at ...

Because every API request now is preceded immediately by a `GET /api/` to determine the `current_version` of the API, _all_ of these tests are now broken.  So this change makes it so that the version determination requests are not recorded in `FauxAdapter.requests` (so our tests don't have to all be changed).

tower_cli/api.py
```diff
@@ -262,9 +266,27 @@ def test_mode(self):
...
+                    if request.url.endswith('/api/'):
+                        self.requests.pop(-1)
+                    return ret
+
+                def clear(self):
+                    super(VersionedFauxAdapter, self).clear()
+                    self._url_pattern = '%s'
+                    self.register_json('https://%s/api/' % settings.host, {
```
ryanpetrello: This registers - by default - a fake response for `/api/` so that version requests within tests succeed.  For now, I've left this hard-coded as `/api/v1/`, because that's what all of our tests expect.  In the future, as the stable version of the Tower API is incremented, we'll want to come in and change this value (and the `/api/v1/` strings hard-coded everywhere in our tests).

tower_cli/api.py
```diff
@@ -262,9 +266,27 @@ def test_mode(self):
                                      password='This is the best wine.',
                                      verbose=False, format='json'):
             adapters = copy.copy(self.adapters)
-            faux_adapter = FauxAdapter(
-                url_pattern=self.prefix.rstrip('/') + '%s',
-            )
+
+            class VersionedFauxAdapter(FauxAdapter):
```
jangsutsr: +1 here. I think this method does minimize the impact of mutable API version on our test suit.

ryanpetrello:
Closing this, sounds like we don't want to go down this route.

tests/test_api.py
jangsutsr: Since `_make_request` is a requests private method, maybe we should wrap it with our own method, so there will be one single place in the code base to be modified in the future?

PR #317, opened 2017-07-06
Categorize help text options

jangsutsr:
This PR increases readability of action command help text by categorizing options into 4 categories.

1. 'Field Options', which are options derived from resource fields.
2. 'Local Options', which are custom options specific to individual action commands.
3. 'Global Options', which are options that control runtime settings keyword arguments.
4. 'Other Options', which includes and only includes `--help` for now.

### DIFF
--- a/tower_cli/cli/action.py
+++ b/tower_cli/cli/action.py
@@ -14,6 +14,9 @@
 # limitations under the License.
 
 import click
+from click.formatting import join_options
+
+from tower_cli.conf import SETTINGS_PARMS
 
 
 class ActionSubcommand(click.Command):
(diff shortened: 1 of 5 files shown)


tower_cli/cli/resource.py
```diff
@@ -340,15 +339,13 @@ def get_command(self, ctx, name):
...
-            new_method)
+        cmd = click.command(name=name, cls=ActionSubcommand, **attrs)(new_method)
 
         # If this method has a `pk` positional argument,
         # then add a click argument for it.
         code = six.get_function_code(method)
         if 'pk' in code.co_varnames:
-            click.argument('pk', nargs=1, required=False,
-                           type=str, metavar='[ID]')(cmd)
+            click.argument('pk', nargs=1, required=False, type=str, metavar='[ID]')(cmd)
```
AlanCoding: I see that you bought yourself some extra line length by moving this into a new file 😉

tower_cli/cli/action.py
```diff
@@ -14,6 +14,9 @@
 # limitations under the License.
 
 import click
+from click.formatting import join_options
+
+from tower_cli.conf import SETTINGS_PARMS
```
AlanCoding: This is a really slick way to implement this feature. I was wondering how you got this working with so little code...

tower_cli/cli/action.py
```diff
@@ -41,3 +44,40 @@ def parse_args(self, ctx, args):
             click.echo(ctx.get_help())
             ctx.exit()
         return super(ActionSubcommand, self).parse_args(ctx, args)
+
+    def format_options(self, ctx, formatter):
```
AlanCoding: So I take it that this is called by the click library? A tiny comment noting that might be nice.

tower_cli/cli/resource.py
jangsutsr: Yup, let's migrate to 120c/line world bit by bit~

tower_cli/cli/action.py
jangsutsr: Will do.

Issue #284, opened 2017-03-27
workflow schema does not replace existing workflow


wenottingham:
Maybe I'm misunderstanding, but I have an existing workflow like this:

```
- job_template: 12
- job_template: 12
```

I have a schema file that defines the following structure:

```
- job_template: 13
  success_nodes:
  - failure_nodes:
    - job_template: 14
    job_template: 15
  - failure_nodes:
    - job_template: 14
    job_template: 12
```

I apply it, and I expect it to be as defined in the schema. Instead I get:

```
- job_template: 12
- job_template: 12
- job_template: 13
  success_nodes:
  - failure_nodes:
    - job_template: 14
    job_template: 15
  - failure_nodes:
    - job_template: 14
    job_template: 12
```

AlanCoding:
That was kind of intended [docs](https://github.com/ansible/tower-cli/blob/master/docs/WORKFLOWS.md):

> The schema command can not delete nodes, so running multiple commands with different schemas will result in creating multiple branches and/or sub-branches.

When looping over the base nodes, it checks to see if a node using JT pk=13 exists. Apparently not finding such a node in that workflow, it creates a new one and carries on.

To do what you expected, delete the workflow, create a new one, and run the schema command. We could certainly have a flag like `--destroy-existing` that does this same thing. I don't want to get into the business of hand-picking ones to delete. If the user wants to delete a node, they should somehow go about starting over.

wenottingham:
That is different than how other commands work; I think it would make more sense to have it follow how we do other resource creation/editing.

AlanCoding:
> I think it would make more sense to have it follow how we do other resource creation/editing.

We do that with how the nodes work, but the problem with that system is that nodes don't have any named accessor that you can use. You have to save and store the pk. Nodes are unique in this way because of the Tower implementation.

Extrapolating a bit, but it sounds like what you want is an option `--deleting-existing=true`, which if set, will delete all the existing nodes in the WFJT before creating the data in the provided schema. It sounds like you might also want that to be on by default.

wenottingham:
Yes, that would work for me. I do believe that even if we end up deciding not to change the default we definitely want that option.

PR #322, opened 2017-07-10
Enable workflow schema update.

jangsutsr:
Connect #284.

See updates in `WORKFLOWS.md` for behavior changes.

### DIFF
--- a/docs/WORKFLOWS.md
+++ b/docs/WORKFLOWS.md
@@ -66,9 +66,9 @@ tower-cli workflow schema workflow1
 
 Here, "workflow1" is the name of the workflow.
 
-### Creating a Schema Definition
+### Creating/updating a Schema Definition
 
-To bulk-create a workflow node network, use the workflow schema command.
+To bulk-create or buld-update a workflow node network, use the workflow schema command.
 The schema is JSON or YAML content, and can be passed in the CLI
 argument, or pointed to a file. The schema is passed as a second positional
 argument, where the first argument references the workflow.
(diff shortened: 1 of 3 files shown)


AlanCoding:
It looks like the node ids are returned with a simple schema GET,

```
$ tower-cli workflow schema asdf
- id: 1
  job_template: 18
```

I see why that might be desired to add, but I'm wondering if it was intended.

jangsutsr:
@AlanCoding No it's not. I just don't want to explicitly add that handling and increase code confuse.

AlanCoding:
Oh yeah, that's totally fine. I will just want to check that creation of a new WFJT will work the same given the schema output.

AlanCoding:
This looks excellent with my testing so far. Here are some materials that I'm gathering for this, along with the other feature about looking up inventory sources by a partial name:

https://github.com/AlanCoding/tower-cli/commit/4096c6db4e0dd304228365db7e12d31d0292aa40

I've been testing switching between two different schemas, and it's going very well. When telling it to apply the schema example after it already exists, I see 8 GET requests and no changes made. That is _nice_ idempotency.

I still might want to cycle through one more example early tomorrow with a more complex branching structure. So far this is incredible.

AlanCoding:
The culmination of my testing:

Schema A

```yaml
# Branch that has deeply nested content removed  going to schema B
- job_template: workflow JT
  failure_nodes:
    - inventory_source: tower script1 (workflow demo
  success_nodes:
  - project: Ansible Examples
    always_nodes:
    - job_template: workflow JT
      success_nodes:
      - job_template: workflow JT

# Branch that is modified going to schema B
- job_template: workflow JT2
  failure_nodes:
  - job_template: workflow JT2
  - job_template: workflow JT
  success_nodes:
  - job_template: workflow JT2
  - job_template: workflow JT

# Branch that is deleted going to schema B
- project: Ansible Examples
  always_nodes:
  - inventory_source: tower script1 (workflow demo
  - inventory_source: tower script2 (workflow demo
```

Schema B

```yaml
# Branch that has deeply nested content removed  going to schema B
- job_template: workflow JT
  failure_nodes:
    - inventory_source: tower script1 (workflow demo
  success_nodes:
  - project: Ansible Examples
    always_nodes:
    - job_template: workflow JT

# Branch that is modified going to schema B
- job_template: workflow JT2
  failure_nodes:
  - inventory_source: tower script1 (workflow demo
  - inventory_source: tower script2 (workflow demo
  success_nodes:
  - project: Ansible Examples
  - job_template: workflow JT
```

This feature seems to be able to swap back and fourth between A and B. I see DELETEs and POSTs right where I expect them. Reapply either A or B after that state has already been reached, and it makes only GET requests, no changes are done.

I have actually avoided the indeterminate cases where 2 of the same JTs are on the same level next to each other. I presume that you have tested this, but I don't think it's high enough customer value to bother with right now. I would recommend that schema authors just avoid that, and my example schemas will, thus, follow the rule as well.

Also, we can't really have any PATCH actions with the approach to schema authoring I'm testing. I presume we would if the ids were used, but I would also hesitate to directly suggest that method as a reusable and sharable format.

This is tremendously complex, and executed perfectly.

PR #328, opened 2017-07-18
Usage compatibility fix for Tower modules

jangsutsr:
To make sure Tower modules' usage (like [here](https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/web_infrastructure/ansible_tower/tower_job_cancel.py#L69)) does not cause import errors.


AlanCoding:
the uses I see:

 - [x] tower_cli.get_resource
 - [x] tower_cli.conf.settings

looks good

PR #336, opened 2017-08-09
allow for multiple inheritances with click methods

AlanCoding:
This has caused bugs in the past. I hope the test is expressive enough to show what is being solved.

Basically, we previously couldn't use mixins, because the click decoration would not be respected across multiple parent classes (see the naive `superclass = bases[0]`), which means that almost any attempt at using multiple classes would result in missing or incomplete commands.

Note that python functionality was never impacted. That was never really a problem. This deals with `@command(**kwargs)` in parent classes, mixins, and subclasses - how to choose which kwargs and other args to use and such.

### DIFF
--- a/tests/test_models_base.py
+++ b/tests/test_models_base.py
@@ -90,8 +90,46 @@ class ResourceMetaTests(unittest.TestCase):
 
         # Establish that it has one of the options added to the
         # superclass list command.
-        self.assertTrue(any([i.name == 'query'
-                            for i in MyResource.list.__click_params__]))
+        self.assertEqual(MyResource.list.__click_params__, models.Resource.list.__click_params__)
+
+    def test_multiple_inheritance(self):
+        """
+        Establish that click decoration from all parent class chains are
+        preserved in a subclass.
+        """
+        class MyMixin(models.Resource):
+            endpoint = '/bogus/'
+
+            def list(self, **kwargs):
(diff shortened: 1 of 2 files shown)


AlanCoding:
One option that might be a good alternative for how to manage this: if `_cli_command_attrs` is not present, raise a descriptive error message that methods across all classes in the inheritance chain must be consistently decorated with the command decorator.

For instance, if the subclass defines `list`, and they don't decorate it, they'll just get a message telling them to add that.

I don't _think_ there's any good argument one way or the other.

Later, I want to come back to test combining over many classes, and assure that arguments are not duplicated.

jangsutsr:
I'm good with this PR as long as arguments are verified to have no duplicates (which is the case from code review). Please don't forget to duplicate and transplant this branch onto master (in our internal repo for now) and open a linked PR there before merging this PR (That would be our normal devel workflow from now on)

AlanCoding:
@jangsutsr good catch on the duplicate options. Added a test case for that and fixed it.

AlanCoding:
I went back over this and gave it a little more careful consideration regarding the precedence. For example:

`class Resource(Parent1, Parent2):`

In this case, Parent1 should win out over Parent2, if they both define the same thing with different values. Of course, the class itself should be the ultimate winner. I re-phrased some of the code to do passive updating along the precedence order.

PR #349, opened 2017-09-02
new API-version-specific install mechanism

AlanCoding:
This introduces a new means of doing a parallel install of tower-cli on your system (will document as a "secondary install") that exists independently. Use by running `sudo make v1`.

The usefulness of such a secondary install is that you can continue to alter what we will call the "primary install" without any changes being made to the secondary install. Use:

```
$ tower-cli-v1
Usage: tower-cli-v1 [OPTIONS] COMMAND [ARGS]...

Options:
  --version  Display tower-cli-v1 version.
  --help     Show this message and exit.

Commands:
  ad_hoc                 Launch commands based on playbook given at...
  config                 Read or write tower-cli-v1 configuration.
  credential             Manage credentials within Ansible Tower.
  group                  Manage groups belonging to an inventory.
  host                   Manage hosts belonging to a group within an...
  inventory              Manage inventory within Ansible Tower.
  inventory_script       Manage inventory scripts within Ansible...
  job                    Launch or monitor jobs.
  job_template           Manage job templates.
  label                  Manage labels within Ansible Tower.
  node                   Manage nodes inside of a workflow job...
  notification_template  Manage notification templates within Ansible...
  organization           Manage organizations within Ansible Tower.
  permission             Manage permissions within Ansible Tower in...
  project                Manage projects within Ansible Tower.
  role                   Add and remove users/teams from roles.
  schedule               Manage schedules within Ansible Tower.
  setting                Manage settings within Ansible Tower.
  team                   Manage teams within Ansible Tower.
  user                   Manage users within Ansible Tower.
  version                Display version information.
  workflow               Manage workflow job templates.
  workflow_job           Launch or monitor workflow jobs.
```

This minimizes the size of the change done in the source-tree by wholesale copying the entire source code folder to a new directory, and renaming _everything_ including the python import path. You can use the secondary install inside of python via `import tower_cli_v1`.

Basic testing shows functionality is in-place. I'm not a huge fan of the need for a `bin_v1` folder, but I don't have any great alternative.


ghjm:
Why do you need sudo privileges to run make?

Also, can you explain the makefile logic? What does version_swap.py do?

AlanCoding:
version_swap.py recursively finds all occurrences of `tower_cli` and replaces them with `tower_cli_v1`. 99% of this is changing the `import tower_cli` lines. In order to keep these two installs separate (and not using different virtualenvs) they need to have their own import paths.

And I don't like the sudo for installing either. Something wasn't done right at some point, but I do not know what. That is a general problem for the installer (issues welcome).

ghjm:
Oh I get it, `make v1` is the v1 equivalent of `make install` which you would expect to require sudo.

The existing `make install` removes the `dist/` and `build/` folders. Do we not also need to do this on v1?

Also, the dependency from `v1` to `clean_v1` is problematic. Suppose you edit a file in the v1 folders and then do (what you think is) a reinstall of v1. It's not polite for the Makefile to unexpectedly `rm -rf` your changes, which for all we know you spent three days working on.

The way I would approach this is to split the current `make v1` target into two - `make prep_v1` and `make install_v1`. Have `prep_v1` copy the folders and create a flagfile (similar to what we do with `.exists` elsewhere in the Makefile). Then have `install_v1` depend on the flagfile.

Though it's possible I'm still missing an essential point here. What do you think?

AlanCoding:
Totally understand where you're coming from with `prep_v1`. That will be the action which lays down the source tree. I think we would still want `clean_v1`, even if we don't pull it into another target. So I'm looking splitting into 3 targets

 - clean_v1
 - prep_v1
 - install_v1

ghjm:
Sounds good to me. Sorry if I wasn't clear - I didn't mean to suggest removing the `clean_v1` target - I just want to make sure it doesn't take users by surprise.

AlanCoding:
I still really want to have a combined target, so I introduced `v1-refresh`, which just combines the other targets.

I also really want to avoid breaking pattern with the existing install target.

Current version of PR is my answer to these concerns, and I did some basic re-testing functionality, submitting for re-review.

ghjm:
I would suggest something like this:

```
setup_v1.py:
	cp -R tower_cli tower_cli_v1/
	cp bin/tower-cli bin/tower-cli-v1
	cp setup.py setup_v1.py
	python version_swap.py

prep_v1: setup_v1.py

install_v1: setup_v1.py
	sudo rm -rf dist/ build/
	sudo python setup_v1.py install
```
The idea here is that if you do `make install_v1` in a fresh checkout, it will do the prep steps rather than just failing. If you do it a second time, it will notice the existence of `setup_v1.py` and avoid re-executing the prep steps.

`v1-refresh` looks fine to me.

AlanCoding:
I discovered a very new kind of bug testing this

```
$ tower-cli-v1
  File "/Users/alancoding/.virtualenvs/cli2/bin/tower-cli-v1", line 4, in <module>
    __import__('pkg_resources').run_script('ansible-tower-cli-v1==3.1.7', 'tower-cli-v1')
  File "/Users/alancoding/.virtualenvs/cli2/lib/python2.7/site-packages/pkg_resources/__init__.py", line 742, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/Users/alancoding/.virtualenvs/cli2/lib/python2.7/site-packages/pkg_resources/__init__.py", line 1503, in run_script
    exec(code, namespace, namespace)
  File "/Users/alancoding/.virtualenvs/cli2/lib/python2.7/site-packages/ansible_tower_cli_v1-3.1.7-py2.7.egg/EGG-INFO/scripts/tower-cli-v1", line 19, in <module>
    from tower_cli_v1.cli.base import TowerCLI
  File "/Users/alancoding/.virtualenvs/cli2/lib/python2.7/site-packages/ansible_tower_cli-3.1.7-py2.7.egg/tower_cli_v1/__init__.py", line 22, in <module>
    __version__ = open('%s/VERSION' % whereami, 'r').read().strip()
IOError: [Errno 2] No such file or directory: u'/Users/alancoding/.virtualenvs/cli2/lib/python2.7/site-packages/ansible_tower_cli-3.1.7-py2.7.egg/tower_cli_v1/VERSION'
```

This is only an issue when installing tower-cli and tower-cli-v1 at the same time with the same versions.

Somehow, the fact that our source used a symlink caused it to be disappear when something happened to the source file. I removed the symlink and moved the version file into the library directory itself - which seems much more durable overall.

Issue #283, opened 2017-03-27
workflow schema output doesn't match other things


wenottingham:
```
[notting@nostromo: ~]$ tower-cli workflow schema testwf @ymlfu
- job_template: 12
- job_template: 12
- job_template: 13
  success_nodes:
  - failure_nodes:
    - job_template: 14
    job_template: 15
  - failure_nodes:
    - job_template: 14
    job_template: 12
```

wenottingham:
The ask is to have this more like the 'normal' ok/changed/failed output.

AlanCoding:
I struggled getting a grasp on this issue, but I think I we can make something work.

### Status Quo

We simply _don't do human format_ for workflow schema commands. We force-switch the runtime format to JSON. Human format will not work because it expects a resource with columns, and this doesn't fit that pattern.

### Proposal

Printing the "changed" status implies going through the human formatting code. That means that we will need a special clause in the human formatting method to echo the contents in YAML (or in the future, perhaps a fancy graphical view).

Similar to other commands, this _also_ implies that the "changed" key will be present for YAML or JSON output, example:

```yaml
- job_template: 12
- job_template: 12
- job_template: 13
  success_nodes:
  - failure_nodes:
    - job_template: 14
    job_template: 15
  - failure_nodes:
    - job_template: 14
    job_template: 12
- changed: True
```

Of course, the changed status should be smart as to whether any unsafe methods (POST, DELETE, etc.) were used in our new fancy idempotent algorithm.

AlanCoding:
We should also robustly produce the "changed" value before creating the `tower_workflow` module for Ansible. FYI @wwitzel3

AlanCoding:
Kicking out to 3.2.1, still don't know the answer to some questions.

AlanCoding:
Closing related PR #350 because I don't know what an acceptable solution could even look like, or maybe the fundamentals of the workflow resources simply haven't been communicated.

PR #350, opened 2017-09-05
[WIP] carry through workflow changed info

AlanCoding:
Connect #283

This gets some of the basics of supporting the `changed` attribute on the data. We already had this data, but sometimes our code abandoned it. This PR is just making sure that we keep accounting for this flag.

Still kind of risky, and I would not want to block 3.1.8 with this. Also, we can't do it without changing the structure of the returned output, because the normal structure has _list_ as the top-level data structure...

screen shot 2017-09-05 at 9 19 25 am

Here I delete one of the branches:

screen shot 2017-09-05 at 9 23 06 am

If you apply the schema with a deleted branch 1 more time, then it's green, of course.

Relevant: https://www.youtube.com/watch?v=pl3vxEudif8


AlanCoding:
I blame github for not processing the git history normally.

PR #366, opened 2017-09-26
Enable association and disassociation of extra_credentials with job_templates

ethackal:
This PR is against tag/v3.1.8, since releases don't appear to be fully merged with master.

Borrowing the basic logic of associating labels to job templates, this PR adds the ability to associate and disassociate extra credentials beyond the initial machine credential for a job template.


AlanCoding:
The versioning idea is that `master` supports v2 API and branch `v1` supports the v1 API.

`extra_credentials` are only really supported through the v2 API, so we will want this in the `master` branch (as you've done here). The only potential problem is that this was branched off of the v1 branch.

I'm going to merge the v1 branch into master and then check back on what git history this PR shows.

AlanCoding:
That didn't quite work as I hoped it would.

Could you try just re-opening this pull request? Other branches forked from the same point have no trouble when I try it, so I suspect something specific to the PR's history is at work here.

if that doesn't work, you could checkout the master branch and then `git cherry-pick 4076b53d93dda39f49f5cdae55e2465f5a27f40b`, and make a new PR from that.

ethackal:
I originally coded off master but then copied to v1. This would be easier if I just go back to my original branch. I'll close this PR and resubmit.

PR #367, opened 2017-09-26
remove old version logic not needed in v2 API

### LINKED ISSUE #358: [v2] Remove special handling for pre-3.0 credentials
The Credential resource has a bunch of special behavior to accommodate old version switches.

I think this is because `user` and `team` used to be genuine fields, but these columns were removed in Tower 3.0. None of this is needed for post-3.2 tower-cli

AlanCoding:
Resolves #358

Some basic demoing:

```
$ tower-cli credential create --credential-type="Source Control" --inputs='{"username": "asdf", "password": "fdsa"}' --user=admin --name='asdf' -v
*** DETAILS: The credential_type field is given as a name; looking it up. *****
GET http://localhost:8013/api/v2/credential_types/
Params: {'name': 'Source Control'}

*** DETAILS: The user field is given as a name; looking it up. ****************
GET http://localhost:8013/api/v2/users/
Params: {'username': 'admin'}

GET http://localhost:8013/api/v2/credentials/
Params: {'name': u'asdf'}

POST http://localhost:8013/api/v2/credentials/
Data: {'inputs': {'username': 'asdf', 'password': 'fdsa'}, 'credential_type': 2, 'user': 1, 'name': u'asdf'}

Resource changed.
== ==== =============== 
id name credential_type 
== ==== =============== 
 8 asdf               2
== ==== =============== 
```

We might have some other bugs to file - it's not using credential type in the lookup, not providing name doesn't raise an error right away.

### DIFF
--- a/tests/test_resources_credential.py
+++ b/tests/test_resources_credential.py
@@ -73,12 +73,11 @@ class CredentialTests(unittest.TestCase):
             self.assertTrue(cred_res.fields[3].no_lookup)
 
             # Verify request data is correct
-            self.assertEqual(len(t.requests), 3)
-            self.assertEqual(t.requests[0].method, 'OPTIONS')
-            self.assertEqual(t.requests[1].method, 'GET')
-            self.assertEqual(t.requests[2].method, 'POST')
-            self.assertTrue('name=foobar' in t.requests[1].url)
+            self.assertEqual(len(t.requests), 2)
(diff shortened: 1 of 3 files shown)


jangsutsr:
If we no longer have `user` and `team` fields from Tower 3.2, maybe just remove those two TowerCLI fields for good?

AlanCoding:
They are used on creation as a special case still. Even in v2...

PR #372, opened 2017-09-28
[WIP] make API version importable

AlanCoding:
@jangsutsr when I started on this, I missed the fact that you added the version number to `constants.py`. We could switch to that, I would just need some extremely quick method to exact that out of the file using basic python logic.

Objective:

allow other libraries to `from tower_cli import __api__version__`

The idea is that we would ship this in both 3.1.9 as well as 3.2.0, and that users can then leverage it themselves.


AlanCoding:
And I also realize that, with this, we will be transitioning to the cherry-picking branch management strategy for v1 and master.

AlanCoding:
Also note

```
$ tower-cli version
Tower CLI 3.1.8
API version v1
Ansible Tower 3.1.0
```

AlanCoding:
Going to close this and make another one, more similar to https://github.com/ansible/tower-cli/pull/410

PR #381, opened 2017-09-28
Audit changes needed in 3.2 for job resources

jangsutsr:
Audit against OPTIONS return of Tower 3.2 and make up any missing fields
for resources job_template, job, workflow and ad_hoc_command.

Signed-off-by: Aaron Tan <jangsutsr@gmail.com>

### DIFF
--- a/tower_cli/resources/ad_hoc.py
+++ b/tower_cli/resources/ad_hoc.py
@@ -43,18 +43,11 @@ class Resource(models.ExeResource):
         type=click.Choice(['run', 'check']),
     )
     inventory = models.Field(type=types.Related('inventory'))
-    machine_credential = models.Field(
-        'credential',
-        display=False,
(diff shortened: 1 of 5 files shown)


tower_cli/resources/ad_hoc.py
```diff
@@ -85,11 +79,7 @@ class Resource(models.ExeResource):
                   help='If provided with --monitor, this attempt'
                        ' will time out after the given number of seconds. '
                        'Does nothing if --monitor is not sent.')
-    @click.option('--become', required=False, is_flag=True,
-                  help='If used, privilege escalation will be enabled for '
-                       'this command.')
```
AlanCoding: It looks like maybe we aliased `become_enabled` to `become`? I'd say should go in the changelog, but I have no objections.

Issue #343, opened 2017-08-22
Not able to delete project_update and inventory_update job using tower-cli


jainnikhil30:
Feature Request:

Currently we have a way to delete the jobs which were playbook runs using 

```
tower-cli job delete <job-id>
```

but there is no way for us to delete the project_update and inventory_update job types via tower-cli

AlanCoding:
Coming soon in 3.2.0 release

PR #384, opened 2017-09-29
add resources for project/inv updates

AlanCoding:
Connect #343

This isn't perfect, but it will help a lot.

I've tested this with the stdout, delete, and cancel commands. Before, we had no access to these, and all of the logic already works.

I also changed how we are returning responses for basic project and inventory updates. For job launches we return the id, so I want to be consistent with that. But of course, launching a job from the job command has a congruency that launching from the template does not. So it's still weird, but we should be returning the job id on every launch as a best-possible practice.

### DIFF
--- a/tower_cli/resources/inventory_update.py
+++ b/tower_cli/resources/inventory_update.py
@@ -1,0 +1,46 @@
+# Copyright 2017, Red Hat, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


tower_cli/resources/inventory_update.py
```diff
@@ -0,0 +1,47 @@
+# Copyright 2015, Ansible, Inc.
```
AlanCoding: > Copyright 2015, Ansible, Inc.

Whoops, don't want to leave that.

PR #394, opened 2017-10-08
make combine_files() work with UTF-8 encoded files

evgeni:
setup.py might fail on Python 3 in non-UTF8 environments while parsing
HISTORY.rst or README.rst which are not required to be ASCII-only with:

    UnicodeDecodeError: 'ascii' codec can't decode …

We could also require the files to be plain ASCII, but it sounds not the
right thing to do in the 21st century.

Simple reproducer:

    % LC_ALL=C python3 setup.py
      File "setup.py", line 126, in <module>
        long_description=combine_files('README.rst', 'HISTORY.rst'),
      File "setup.py", line 112, in combine_files
        file_contents.append(f.read())
      File "/usr/lib/python3.5/encodings/ascii.py", line 26, in decode
        return codecs.ascii_decode(input, self.errors)[0]
    UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 2132: ordinal not in range(128)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,6 +18,7 @@
 
 import re
 import sys
+import codecs
 from distutils.core import setup
 from setuptools import find_packages
 from setuptools.command.test import test as TestCommand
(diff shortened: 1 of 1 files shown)


AlanCoding:
This makes sense to me. I believe the idea is that files that we are providing are implicitly utf-8, and it just needs to be marked as such.

```
file -I HISTORY.rst 
HISTORY.rst: text/plain; charset=utf-8
```

PR #395, opened 2017-10-08
Travis updates

evgeni:


.travis.yml
```diff
@@ -14,7 +15,7 @@ install:
     - pip install tox-travis
     - pip install coveralls
 before_script:
-    if [[ $TRAVIS_PYTHON_VERSION != '2.6' ]]; then flake8 . --max-line-length=120; fi
+    flake8 . --max-line-length=120
```
AlanCoding: oh right, because we got rid of 2.6

evgeni: Yupp. And less code is better code, right?

PR #396, opened 2017-10-08
fix spelling mistakes spotted by Debian's lintian tool

evgeni:


AlanCoding:
@jangsutsr you'll need to get the docs site source rebuilt, I believe.

PR #412, opened 2017-10-18
make CUR_API_VERSION importable in v1 branch

AlanCoding:
replaces #372


AlanCoding:
I'm going to have to combine this with some other stuff

PR #416, opened 2017-10-23
Get rid of VERSION file and use entry_point in v1 branch

AlanCoding:
In short, vendoring this VERSION file was not playing
nice with setuptools
* make CUR_API_VERSION importable in v1 branch as constant
* change setup and __init__.py code to adjust

Yes, this does need to be adapted to the master branch / v2


AlanCoding:
I thought it might be best to just go all the way to parity with the branches, so I put in the change to get rid of the bin directory here as well.

PR #419, opened 2017-10-24
Get rid of VERSION file in main branch

AlanCoding:
This also has cherry-picked a commit to get rid of the use of `whereami`, which is replaced by pkgutil.


jangsutsr:
Lgtm. Once the documentation for the purpose of this PR ---- advanced installation via `make install_v2` (or whatever) is in, I think it's ready to merge.

setup.py
```diff
@@ -116,10 +117,18 @@ def combine_files(*args):
     return "\n\n".join(file_contents)
 
 
+# Read the constants, for versioning information
+constants = {}
+exec(
+    open(os.path.join(pkg_name, 'constants.py')).read(),
+    constants
+)
```
AlanCoding: @ghjm is this too insecure to use `exec` here, or is it alright because the Makefile forces the relative path, and we already know what the (static) content is? I really don't want to write logic to parse a python file as `key = value` syntax.

ghjm: If you're concerned that the relative path might not work (or might be hijackable) if the current directory is something other than the location of setup.py, you could do something like 
```
exec(
    open(os.path.join(os.path.dirname(os.path.abspath(__file__)), pkg_name, 'constants.py')).read(),
    constants
)
```
but someone will likely come along and say it's overkill.

AlanCoding: Thanks that is a good suggestion. I'll keep this in the back of my head, but I think I'll move forward with the current state of this.

PR #422, opened 2017-10-26
[WIP] explicitly declare package to be zip_safe

AlanCoding:
The install is slightly faster, about 0.7s vs 0.9s with the flag True. I haven't tested the command times.

We might also need to have a special debug mode, if this won't give descriptive tracebacks, but not sure. Needs testing.


AlanCoding:
juggling various versions locally, I have found this to be very, very, dangerous.

Issue #296, opened 2017-05-01
config via environment variable broken


wenottingham:
```
[notting@nostromo: ~]$ tower-cli --version
Tower CLI 3.1.3
[notting@nostromo: ~]$ TOWER_HOST=foo.bar TOWER_USERNAME=cow TOWER_PASSWORD=foo TOWER_VERIFY_SSL=false tower-cli config

# Defaults.
username: 
description_on: False
host: 127.0.0.1
verbose: False
certificate: 
format: human
color: True
password: 
verify_ssl: True
```

Have not dug through to see when this last worked as documented.

AlanCoding:
Tried with 3.0.2 and replicated with the steps you gave. The content it echos is unchanged by the env variables. Ping @wwitzel3 so he knows.

jangsutsr:
The problem here is that documentation is not accurate enough. Because environment variables won't be included when using Tower CLI as a command-line tool. Under the hood, env variables only affect `setting.runtime_values` context manager, and that context manager is only used by tests and  ansible tower modules.

Will update docs to remove the confusion.

wenottingham:
Is there any reason *not* to support it when using the CLI?

jangsutsr:
@wenottingham Because reading configuration from env variables has not been taken into consideration before, but we definitely can support it. Ping @AlanCoding for opinion on adding this feature.

mauricios:
This feature is very useful to use Tower CLI in automation tools. I use the environment variables to change Tower CLI configuration programmatically from Hubot scripts, for example changing the user, password and event hostname of Tower CLI depending on the chat user or task request to Hubot from Slack.

PR #424, opened 2017-10-30
always read settings from environment

AlanCoding:
Connect #296 

```
$ TOWER_HOST=foo.invalid tower-cli user list
Error: There was a network error of some kind trying to connect to Tower.

The most common  reason for this is a settings issue; is your "host" value in `tower-cli config` correct?
Right now it is: "foo.invalid".
```

### DIFF
--- a/tests/test_conf.py
+++ b/tests/test_conf.py
@@ -112,25 +112,23 @@ class ConfigFromEnvironmentTests(unittest.TestCase):
     def test_no_override(self):
         """Establish that the environment variables do not override explicitly
         passed in values."""
-        settings = Settings()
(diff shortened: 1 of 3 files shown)


AlanCoding:
One potentially small problem: this could make the behavior of test pass/fail dependent on the environment variables.

I don't want to worry about that right now, because we should be coding for the usage, not for the tests.

AlanCoding:
just realized -

this will need some extra work to show the environment variables when we echo the config `tower-cli config`

AlanCoding:
got that working now. It's pretty cool.

```
 tower-cli config

# Options from environment variables.
host: foo.invalid
```

PR #437, opened 2017-11-08
Revert "updating tower-cli docs to use the standard ansible tower themes and …"

jangsutsr:
Reverts ansible/tower-cli#411

Since we don't have plan to actually include CLI docs to official site, this PR should be reverted as it blocks readthedocs rendering.


AlanCoding:
there's a strange flake8 failure, maybe due to the branch point?

PR #441, opened 2017-11-21
Updates for saved launch-time configurations

AlanCoding:
corresponds to active feature work in AWX

https://github.com/ansible/awx/compare/devel...AlanCoding:configs_rebased3?expand=1

### DIFF
--- a/tower_cli/resources/node.py
+++ b/tower_cli/resources/node.py
@@ -44,14 +44,33 @@ class Resource(models.Resource):
     workflow_job_template = models.Field(
         key='-W', type=types.Related('workflow'))
     unified_job_template = models.Field(required=False)
+
+    # Prompts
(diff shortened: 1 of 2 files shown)


tower_cli/resources/job_template.py
```diff
@@ -139,7 +139,11 @@ def associate_credential(self, job_template, credential):
 
         =====API DOCS=====
         """
-        return self._assoc('extra_credentials', job_template, credential)
+        try:
+            # Pattern in 3.3
+            return self._assoc('credentials', job_template, credential)
```
AlanCoding: might want `exc. NotFound` for the exception, also might want to log this.


But we almost might prefer a new exception type for #442

AlanCoding:
I reduced the scope, this is just tracking new fields now. That should make it easier to work with.

Issue #404, opened 2017-10-12
Vault credentials cannot be passed to job launch


viggeh:
Since Ansible Tower 3.2.0, machine credentials now longer include vault encryption information and have been split into machine credentials and vault credentials.

If you have a job template that needs vault credentials (or other types of credentials), it is not possible to pass the vault credentials with ansible-tower-cli. There seems to be a field called `extra_credentials` in the api that could be used for this.

AlanCoding:
I think you want to use the `vault_credential` field for this. You can apply it to your template here:

https://github.com/ansible/tower-cli/blob/master/tower_cli/resources/job_template.py#L42

The current API does not appear to support `ask_vault_credential_on_launch`, so you would not be able to provide it on job launch. However, I think you can have the vault credential ask to prompt you for your password on launch.

AlanCoding:
I was wrong about the ask_ variable, if you set `ask_credential_on_launch` you can provide the vault credential on launch.

We probably want to add that field to the job launch fields.

saurabhrpatil:
Hi,
   I wish to use vault-password-file feature in tower-cli similar to the implementation in ansible-playbook .
ex:  ansible-playbook xyz.yml  -e "host_group=localhost"  **--vault-password-file=/tmp/pass.txt**

 As of today ,we do not have support for  ask_vault_credential_on_launch in  tower-cli  launch template.
i.e in tower-cli job launch --job-template ..

Can you please let us know by when would this feature be available in tower-cli ?

AlanCoding:
marked for 3.3.8 release

PR #449, opened 2017-12-05
[WIP] Allow specifying list of credentials and vault in 3.3

AlanCoding:
Connect #404 

This is actually leap-frogging the development in Tower and AWX, which is possible because we neglected vault-credential support on launch up until now.

If you had Tower 3.2, you weren't able to specify a vault credential on launch. With this PR, it will remain that way (although with a nice error message). If you have AWX, however, this will let you specify vault credential, or multiple `--credential="cred1" --credential="cred2"` directly.

### DIFF
--- a/tower_cli/resources/job.py
+++ b/tower_cli/resources/job.py
@@ -19,7 +19,7 @@ from distutils.version import LooseVersion
 
 import click
 
-from tower_cli import models, get_resource, resources
+from tower_cli import models, get_resource, resources, exceptions as exc
 from tower_cli.api import client
 from tower_cli.cli import types
 from tower_cli.utils import debug, parser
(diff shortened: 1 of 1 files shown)


AlanCoding:
On second thought, the leap-frog might not be a good idea. It's probably better to just pass vault_credential as vault_credential if it's given.

tower_cli/resources/job.py
```diff
@@ -74,8 +74,12 @@ class Resource(models.ExeResource):
     @click.option('--verbosity', type=int, required=False, help='Specify verbosity of the playbook run.')
     @click.option('--inventory', required=False, type=types.Related('inventory'),
                   help='Specify inventory for job template to run.')
-    @click.option('--credential', required=False, type=types.Related('credential'),
-                  help='Specify machine credential for job template to run.')
+    @click.option('--credential', required=False, multiple=True, type=types.Related('credential'),
+                  help='Specify machine credential(s) for job template to run.')
+    @click.option('--vault-credential', required=False, type=types.Related('credential'),
+                  help='Specify vault credential for job template to run.')
+    @click.option('--cloud-credential', required=False, type=types.Related('credential'),
```
ryanpetrello: I'm trying to decide if I want to provide this, or just go all-in on `--credentials`.

AlanCoding: I agree, we should axe cloud & vault here.

However, the `--credential` flag in this PR does what you want it to do, the plural just isn't how click handles multiples. The pattern is `--credential="cred1" --credential="cred2"`.

Issue #406, opened 2017-10-13
job explanation not set when launching a job with tower-cli


anderm3:
When running `tower-cli job launch -vvvv -J 10 --job-explanation foo` I see that the job is launched with the job explanation set however when I get the job information the job_explanation field always comes back as "".

Tested with tower-cli 3.2.0 against an AWX host version 1.0.1.0

AlanCoding:
This should be marked as a read-only field.

It's only there in case someone was to search based on the field. It is only filled non-empty when the job encountered an error, and the field is used to describe what type of error it was.

anderm3:
After some fantastic support I was able to solve the issue by setting the description on the scheduled job. Then later I can list the jobs and use a query string to search the description field for my particular identifier and verify that the job for that given id is already in the system.

```
$ tower-cli schedule list -Q description 42
No records found.
$ tower-cli schedule create -n "Process 42" -d 42 --job-template 10 --rrule "DTSTART:20171017T212500Z RRULE:FREQ=DAILY;INTERVAL=1"
Resource changed.
== ========================= ==================== =======
id           name            unified_job_template enabled
== ========================= ==================== =======
17 Process 42                                  10    true
== ========================= ==================== =======

...

$ tower-cli schedule list -Q description 42
== ========================= ==================== =======
id           name            unified_job_template enabled
== ========================= ==================== =======
17 Process 42                                  10    true
== ========================= ==================== =======
```

PR #450, opened 2017-12-05
make job_explanation read only

AlanCoding:
Connect #406


AlanCoding:
huh, it might be good to make inventory a field, and we could make credentials a field, but it would need some new settings added to the field to make it read-only.

Issue #467, opened 2018-02-09
Divide root-level help text into resources & commands


AlanCoding:
current

```
Usage: tower-cli [OPTIONS] COMMAND [ARGS]...

Options:
  --version  Display tower-cli version.
  --help     Show this message and exit.

Commands:
  ad_hoc                 Launch commands based on playbook given at...
  config                 Read or write tower-cli configuration.
  credential             Manage credentials within Ansible Tower.
  credential_type        Manage credential types within Ansible Tower.
  group                  Manage groups belonging to an inventory.
  host                   Manage hosts belonging to a group within an...
  instance               Check instances within Ansible Tower.
  instance_group         Check instance groups within Ansible Tower.
  inventory              Manage inventory within Ansible Tower.
  inventory_script       Manage inventory scripts within Ansible...
  inventory_source       Manage inventory sources within Ansible...
  inventory_update       Launch or monitor inventory source updates.
  job                    Launch or monitor jobs.
  job_template           Manage job templates.
  label                  Manage labels within Ansible Tower.
  node                   Manage nodes inside of a workflow job...
  notification_template  Manage notification templates within Ansible...
  organization           Manage organizations within Ansible Tower.
  project                Manage projects within Ansible Tower.
  project_update         Launch or monitor project updates.
  role                   Add and remove users/teams from roles.
  schedule               Manage schedules within Ansible Tower.
  setting                Manage settings within Ansible Tower.
  team                   Manage teams within Ansible Tower.
  unified_job            Combined model of projects, job templates,...
  user                   Manage users within Ansible Tower.
  version                Display full version information.
  workflow               Manage workflow job templates.
  workflow_job           Launch or monitor workflow jobs.
```

Proposed

```
Usage: tower-cli [OPTIONS] COMMAND [ARGS]...

Options:
  --version  Display tower-cli version.
  --help     Show this message and exit.

Commands:
  version                Display full version information.
  config                 Read or write tower-cli configuration.

API Resources:
  ad_hoc                 Launch commands based on playbook given at...
  credential             Manage credentials within Ansible Tower.
  credential_type        Manage credential types within Ansible Tower.
  group                  Manage groups belonging to an inventory.
  host                   Manage hosts belonging to a group within an...
  instance               Check instances within Ansible Tower.
  instance_group         Check instance groups within Ansible Tower.
  inventory              Manage inventory within Ansible Tower.
  inventory_script       Manage inventory scripts within Ansible...
[...]

PR #470, opened 2018-02-14
Create new section for commands vs resources

AlanCoding:
#467

This creates the help text that was drafted in the linked issue.

The thinking here is to make it more clear as more global (not resource-specific) commands are added, @john-westcott-iv

This has some code copied from https://github.com/pallets/click/blob/6.x-maintenance/click/core.py, I might need to mark those more clearly.


AlanCoding:
I think the naming could be better. I could see us doing something like

 - Global Commands:
 - API Resources:

PR #473, opened 2018-02-16
Add roles for new AWX feature

AlanCoding:


ryanpetrello:
This looks reasonable to me.  @wwitzel3 do you have any feedback before we merge this?

PR #482, opened 2018-02-28
cache legacy Tower tokens with reasonable permissions

ryanpetrello:


AlanCoding:
https://github.com/ansible/tower-cli/blob/499f4ad39c92f1b619c042ecdf0c559693c7b92d/tower_cli/cli/misc.py#L200

```
>>> (stat.S_IRUSR | stat.S_IWUSR) == 0600
True
```

PR #483, opened 2018-03-02
Make demo script run again

AlanCoding:
This flag was removed, it never worked.


AlanCoding:
Why is coveralls commenting on my PR 4 times for no reason?

Issue #478, opened 2018-02-27
Support managing tokens and applications


AlanCoding:
Implement in tower-cli https://github.com/ansible/awx/pull/904

endpoints under `/api/o/`

```
{
    "authorize": "/api/o/authorize/",
    "token": "/api/o/token/",
    "revoke_token": "/api/o/revoke_token/"
}
```

https://github.com/ansible/awx/blob/devel/docs/auth/oauth.md

ryanpetrello:
@matburt @AlanCoding how far do we think we want to go with this?  At least in terms of immediate priority for tower-cli issues, I'm not sure I want to spend time updating `tower-cli` to deal in the actual creation/management of OAuth apps - there's already other plenty of other tools that can be used to interact with OAuth2 endpoints (including the UI in Tower itself).

Do we want to allow people to provide `user+pass+secret_id+secret_key` for an app and generate a token?  Or do we just want to add a flag for specifying a Personal Auth Token?

In my mind, the minimal path to OAuth2 support here is:

1.  Go into Tower and create your app.
2.  `POST` to `/api/o/token/` to generate a personal access token.
3.  Run `tower-cli --oauth-token=<token-value>`

AlanCoding:
I started this issue posting one thing, but here is a different thing.

Both applications and tokens are present as resources under the `/api/v2/` path. These can be added as ordinary resources. This is dead-simple copy+paste and replace kind of effort.

I need to understand more about the standard and the implementation to make any helpful comment on the `/api/o/` endpoints. If I say something now, I will probably be wrong, so this is the end of my comment.

matburt:
There's a lot we can do here, and I think having the oauth2 implementation in Tower/AWX opens up a world of possibilities.

So... most CLI tools use a workflow that works really well with oauth. Look at oc/kubectl for example. You `oc login` give it your username and password and it goes out and fetches your token... stores that instead of your u/p

gcloud will open a browser and let you authenticate that way and pass the token back to the command line tool... this is pretty nifty-keen also since we support ldap/saml/etc. and I believe our 2-legged system would support this kind of thing

ryanpetrello:
[...]

PR #484, opened 2018-03-06
support authentication via oauth2 personal access tokens

ryanpetrello:

```
$ tower-cli config host tower.example.org
Configuration updated successfully.

$ tower-cli login bob
Password:
Configuration updated successfully.

$ cat ~/.tower_cli.cfg
[general]
host = tower.example.org
oauth_token = 6LvR6QhttzU2cpIBHc2uOqMbCEbPcH

$ tower-cli inventory list
== ============== ============
id      name      organization
== ============== ============
 1 Demo Inventory            1
== ============== ============
```

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 import json
+import warnings
 from datetime import datetime as dt, timedelta
 
 import requests
@@ -43,14 +44,14 @@ class ClientTests(unittest.TestCase):
         URL prefix given a host with no specified protocol.
         """
         with settings.runtime_values(host='33.33.33.33'):
-            self.assertEqual(client.prefix, 'https://33.33.33.33/api/%s/' % CUR_API_VERSION)
+            self.assertEqual(client.get_prefix(), 'https://33.33.33.33/api/%s/' % CUR_API_VERSION)
 
     def test_prefix_explicit_protocol(self):
         """Establish that the prefix property returns the appropriate
         URL prefix and don't clobber over an explicit protocol.
         """
         with settings.runtime_values(host='bogus://33.33.33.33/'):
-            self.assertEqual(client.prefix, 'bogus://33.33.33.33/api/%s/' % CUR_API_VERSION)
+            self.assertEqual(client.get_prefix(), 'bogus://33.33.33.33/api/%s/' % CUR_API_VERSION)
 
     def test_request_ok(self):
         """Establish that a request that returns a valid JSON response
@@ -191,7 +192,7 @@ class ClientTests(unittest.TestCase):
         with settings.runtime_values(
                 host='http://33.33.33.33', verify_ssl=True):
             with self.assertRaises(exc.TowerCLIError):
-                client.prefix
+                client.get_prefix()
 
     def test_failed_suggestion_protocol(self):
         """Establish that if connection fails and protocol not given,
@@ -248,7 +249,9 @@ class TowerAuthTokenTests(unittest.TestCase):
                     t.register('/authtoken/', json.dumps({}), status_code=200, method='OPTIONS')
                     t.register('/authtoken/', json.dumps({'token': 'barfoo', 'expires': expires}), status_code=200,
                                method='POST')
-                    self.auth(self.req)
+                    with warnings.catch_warnings():
+                        warnings.simplefilter("ignore", UserWarning)
+                        self.auth(self.req)
                     self.assertEqual(self.req.headers['Authorization'], 'Token barfoo')
 
     def test_reading_invalid_token_from_server(self):
@@ -269,5 +272,16 @@ class TowerAuthTokenTests(unittest.TestCase):
             with settings.runtime_values(use_token=True):
                 t.register('/authtoken/', json.dumps({}), status_code=404, method='OPTIONS')
                 auth = BasicTowerAuth('alice', 'pass', client)
-                auth(self.req)
+                with warnings.catch_warnings():
+                    warnings.simplefilter("ignore", UserWarning)
+                    auth(self.req)
                 assert self.req.headers == {'Authorization': 'Basic YWxpY2U6cGFzcw=='}
+
+    def test_oauth_bearer_token(self):
+        token = 'Azly3WBiYWeGKfImK25ftpJR1nvn6JABC123'
+        with settings.runtime_values(oauth_token=token):
+            auth = BasicTowerAuth(None, None, client)
+            auth(self.req)
+            assert self.req.headers == {
+                'Authorization': 'Bearer {}'.format(token)
+            }
--- a/tests/test_cli_misc.py
+++ b/tests/test_cli_misc.py
@@ -13,17 +13,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import json
 import os.path
 import stat
 import warnings
 
 import click
 from click.testing import CliRunner
+from fauxquests.response import Resp
 import requests
 
 import tower_cli
-from tower_cli.api import client
-from tower_cli.cli.misc import config, version, _echo_setting
+from tower_cli.api import client, Client
+from tower_cli.cli.misc import config, version, login, _echo_setting
 from tower_cli.conf import settings
 from tower_cli.constants import CUR_API_VERSION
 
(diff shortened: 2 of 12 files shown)


ryanpetrello:
In another PR, I'd like to extend this new `login` command to support app-based token generation, e.g.,

`$ tower-cli login bob --client abc --secret xyz`

tests/test_cli_misc.py
```diff
@@ -269,6 +272,106 @@ def test_error_no_global_config_file(self):
...
+            with mock.patch.object(os, 'chmod'):
+                with client.test_mode as t:
+                    # You have to modify this internal private registry to
+                    # register a URL endpoint that _doesn't_ have the version
+                    # prefix
+                    prefix = Client().get_prefix(include_version=False)
+                    t._registry[six.text_type('{}o/').format(prefix)] = Resp(
+                        '', 200, {}
+                    )
+                    t.register('/users/bob/personal_tokens/', json.dumps({
```
rooftopcellist: I'm curious, is it ok to have 'bob' there instead of the user's pk?

tower_cli/cli/misc.py
```diff
@@ -204,3 +205,30 @@ def config(key=None, value=None, scope='user', global_=False, unset=False):
...
+        '/users/{}/personal_tokens/'.format(username),
+        data={"description": "Tower CLI", "application": None, "scope": "read"},
+        headers={'Authorization': 'Basic %s' % base64.b64encode(':'.join([
+            username, password
+        ]))}
+    )
+
+    if r.status_code == 201:
+        token = r.json()['token']
+        config.main(['oauth_token', token, '--scope=user'])
```
AlanCoding: Well this is a very interesting approach.

AlanCoding:
So is `oauth_token` shown as a global option for all commands? Does it work when used that way? (getting it from the runtime settings)

Can you set it with an env var?

tests/test_cli_misc.py
ryanpetrello: Named URLs :D

ryanpetrello:
I'm super excited about the fact that these tests pass locally for me :)

tower_cli/cli/misc.py
ryanpetrello:
```shell
~/dev/tower-cli tower-cli config

# User options (set with `tower-cli config`; stored in ~/.tower_cli.cfg).
host: tower.example.org:8043

# Defaults.
username:
use_token: False
verbose: False
certificate:
format: human
color: True
description_on: False
verify_ssl: True
oauth_token:
password:

~/dev/tower-cli tower-cli inventory list --tower-oauth-token=PB3X8L2hXjN4mxdYxt7dd0PbInhSDk
== ============== ============
id      name      organization
== ============== ============
 1 Demo Inventory            1
== ============== ============
```

ryanpetrello:
@AlanCoding 

> So is oauth_token shown as a global option for all commands? Does it work when used that way? (getting it from the runtime settings)

With the latest code in this PR:

```shell
~/dev/tower-cli tower-cli config

# User options (set with `tower-cli config`; stored in ~/.tower_cli.cfg).
host: tower.example.org:8043

# Defaults.
username:
use_token: False
verbose: False
certificate:
format: human
color: True
description_on: False
verify_ssl: True
oauth_token:
password:

~/dev/tower-cli tower-cli inventory list --tower-oauth-token=PB3X8L2hXjN4mxdYxt7dd0PbInhSDk
== ============== ============
id      name      organization
== ============== ============
 1 Demo Inventory            1
== ============== ============
```

AlanCoding:
First-world problems: the variables are showing up in defaults even if I haven't set them.

```
$ tower-cli config

# User options (set with `tower-cli config`; stored in ~/.tower_cli.cfg).
username: admin
host: http://localhost:8013/
verify_ssl: False
verbose: False

# Defaults.
use_token: False
certificate: 
format: human
color: True
description_on: False
oauth_token: 
password: 
```

There's got to be some way to recognize a null-ish state, because presumably that would just be copying the implementation for all the other settings.

AlanCoding:
If I can `tower-cli login admin`, then I feel like I should be able to `tower-cli logout admin`

AlanCoding:
It takes an uncomfortably long amount of time after I type my password and press enter for the command to finish. It makes you think that you did something wrong. I think it would be good to give a message like "waiting for carrier pidgins to return"

AlanCoding:
I also don't have a verbose option to debug what requests are being made. We don't have that option for config, but no requests are being made for that.

AlanCoding:
should logging out actually reach out and invalidate the token?

AlanCoding:
the response gives

```json
"expires": "2018-03-07T23:02:15.648450Z", 
```

That would be great information to pass along at the end of the command!

tower_cli/cli/misc.py
```diff
@@ -204,3 +205,30 @@ def config(key=None, value=None, scope='user', global_=False, unset=False):
...
+    """
+    Retrieves and stores an OAuth2 personal auth token.
+    """
+    if not supports_oauth():
+        raise exc.TowerCLIError(
+            'This version of Tower does not support OAuth2.0'
+        )
+
+    # Explicitly set a basic auth header for PAT acquisition (so that we don't
+    # try to auth w/ an existing user+pass or oauth token in a config file)
```
rooftopcellist: We may want to change `oauth` to `oauth2`for consistency with AWX.

rooftopcellist:
@AlanCoding -- Re:
> should logging out actually reach out and invalidate the token?

Can the user continue to use this token to authenticate after changing tower-cli tower A to tower B, then back to A?  

If I understand this right: on `tower-cli login admin`, it passes user/pass to acquire a token, then uses that token for all future tower-cli requests.  If this token is only used for tower-cli, it would be ok to revoke on `tower-cli logout admin`.  

My worry is that a consumer of tower may see the token and try to use it for something else, then be disappointed when the token disappeared from a seemingly unrelated tower-cli logout.  Solution: we could add a warning message on login and logout.  (or just logout)

AlanCoding:
Those are good points, and right now I tend to agree with this behavior

> If this token is only used for tower-cli, it would be ok to revoke on `tower-cli logout admin`.  

In the logout help text, we should offer `tower-cli config oauth_token --unset` as an alternative that keeps the token active.

ryanpetrello:
@AlanCoding @rooftopcellist this PR is getting pretty tall.  I've addressed:

- add a `-v` flag for `tower-cli login`
- print the response JSON (including token value and expiration) when `tower-cli login` succeeds
- added a `--scope` flag, which takes `'read'` and `'write'`, but which defaults to `'write'`

Would it be okay if I added `tower-cli logout` in a separate PR?

rooftopcellist:
I think splitting it up would be fine in this case.  @AlanCoding So to be clear, the --unset flag would just make tower-cli use basic auth, but still keep the token stored, right?

tests/test_api.py
```diff
@@ -248,7 +249,9 @@ def test_reading_expired_token(self):
                     t.register('/authtoken/', json.dumps({}), status_code=200, method='OPTIONS')
                     t.register('/authtoken/', json.dumps({'token': 'barfoo', 'expires': expires}), status_code=200,
                                method='POST')
-                    self.auth(self.req)
+                    with warnings.catch_warnings():
```
ryanpetrello: Tests were noisy without this.

tests/test_api.py
```diff
@@ -269,5 +272,16 @@ def test_auth_token_unsupported(self):
             with settings.runtime_values(use_token=True):
                 t.register('/authtoken/', json.dumps({}), status_code=404, method='OPTIONS')
                 auth = BasicTowerAuth('alice', 'pass', client)
-                auth(self.req)
+                with warnings.catch_warnings():
```
ryanpetrello: Same - noisy tests.

tower_cli/api.py
```diff
@@ -156,8 +166,7 @@ def _make_request(self, method, url, args, kwargs):
                 'Right now it is: "%s".' % settings.host
             )
 
-    @property
-    def prefix(self):
+    def get_prefix(self, include_version=True):
```
ryanpetrello: Up until now, every endpoint we've wanted to hit has been prefixed with a version, so I had to change this from a property to a function for generating `/api/o/` prefixes.

tower_cli/api.py
```diff
@@ -223,12 +236,12 @@ def request(self, method, url, *args, **kwargs):
         # Sanity check: Did we fail to authenticate properly?
         # If so, fail out now; this is always a failure.
         if r.status_code == 401:
-            raise exc.AuthError('Invalid Tower authentication credentials.')
+            raise exc.AuthError('Invalid Tower authentication credentials (HTTP 401).')
```
ryanpetrello: This is beyond the scope of this issue, and PR, but I think it may take debugging a bit clearer for folks (at least it did for me when dealing w/ read vs. write oauth2 scopes).

tower_cli/cli/misc.py
```diff
@@ -204,3 +207,39 @@ def config(key=None, value=None, scope='user', global_=False, unset=False):
...
             )
     click.echo('Configuration updated successfully.')
+
+
+@click.command()
+@click.argument('username', required=True)
+@click.option('--password', required=True, prompt=True, hide_input=True)
+@click.option('--scope', required=False, default='write',
+              type=click.Choice(['read', 'write']))
+@click.option('-v', '--verbose', default=None,
```
ryanpetrello: I opted to just specify `-v` manually instead of inheriting _all_ global options, because some of them don't really make sense for this command.

tower_cli/cli/misc.py
```diff
@@ -204,3 +207,39 @@ def config(key=None, value=None, scope='user', global_=False, unset=False):
...
+    HTTPBasicAuth(username, password)(req)
+    r = client.post(
+        '/users/{}/personal_tokens/'.format(username),
+        data={"description": "Tower CLI", "application": None, "scope": scope},
+        headers=req.headers
+    )
+
+    if r.status_code == 201:
+        result = r.json()
+        result.pop('summary_fields', None)
```
ryanpetrello: These two keys are a bit too verbose/unimportant for this tool.

tower_cli/cli/misc.py
```diff
@@ -204,3 +207,39 @@ def config(key=None, value=None, scope='user', global_=False, unset=False):
...
+        result.pop('related', None)
+        token = result['token']
+        secho(json.dumps(result, indent=1), fg='blue', bold=True)
```
ryanpetrello: screen shot 2018-03-07 at 4 45 52 pm

Issue #401, opened 2017-10-12
Add abstract resource for job events


AlanCoding:
We should create a resource marked as abstract for job events. This can be used for internal purposes, which seems particularly relevant for doing the job event-based standard out.

PR #487, opened 2018-03-12
Add job event resource

AlanCoding:
Fixes https://github.com/ansible/tower-cli/issues/401

### DIFF
--- a/tower_cli/resources/job_event.py
+++ b/tower_cli/resources/job_event.py
@@ -1,0 +1,50 @@
+# Copyright 2018, Red Hat, Inc.
+# Alan Rominger <arominge@redhat.com>
+#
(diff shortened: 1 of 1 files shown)


tower_cli/resources/job_event.py
```diff
@@ -0,0 +1,50 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from tower_cli import models
+from tower_cli.cli import types
+
+
+class Resource(models.BaseResource):
+    """A resource for users."""
```
ryanpetrello: Docstring should be about job events?

PR #490, opened 2018-03-13
support app tokens via `docker login --client-id X --client-secret Y`

ryanpetrello:

### DIFF
--- a/tower_cli/cli/misc.py
+++ b/tower_cli/cli/misc.py
@@ -212,12 +212,14 @@ def config(key=None, value=None, scope='user', global_=False, unset=False):
 @click.command()
 @click.argument('username', required=True)
 @click.option('--password', required=True, prompt=True, hide_input=True)
+@click.option('--client-id', required=False)
+@click.option('--client-secret', required=False)
 @click.option('--scope', required=False, default='write',
               type=click.Choice(['read', 'write']))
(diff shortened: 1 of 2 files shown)


tower_cli/cli/misc.py
```diff
@@ -228,19 +230,37 @@ def login(username, password, scope, verbose):
...
     # try to auth w/ an existing user+pass or oauth2 token in a config file)
+
     req = collections.namedtuple('req', 'headers')({})
-    HTTPBasicAuth(username, password)(req)
-    r = client.post(
-        '/users/{}/personal_tokens/'.format(username),
-        data={"description": "Tower CLI", "application": None, "scope": scope},
-        headers=req.headers
-    )
+    if client_id and client_secret:
```
rooftopcellist: We should also have an `elif` here for if only the `client_id` flag is provided.  In this case, the grant type would be "public".  For example:

```{
    "name": "Public App",
    "description": "for the people",
    "user": 1,
    "client_type": "public",
    "redirect_uris": "",
    "authorization_grant_type": "password",
    "skip_authorization": false
}
```

ryanpetrello: Good point - addressed!

rooftopcellist: LGTM.  Thanks for adding that.

PR #492, opened 2018-03-14
add resources for OAuth2 Application and Token

ryanpetrello:

### DIFF
--- a/tower_cli/resources/token.py
+++ b/tower_cli/resources/token.py
@@ -1,0 +1,34 @@
+# Copyright 2018, Red Hat, Inc.
+# Ryan Petrello <rpetrell@redhat.com>
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


tower_cli/resources/token.py
```diff
@@ -0,0 +1,33 @@
...
+
+    user = models.Field(type=types.Related('user'), required=True)
+    application = models.Field(type=types.Related('application'), required=True)
+
+    created = models.Field(required=False)
+    modified = models.Field(required=False)
+    token = models.Field(required=False)
+    refresh_token = models.Field(required=False)
+    expires = models.Field(required=False)
+    scope = models.Field(required=False)
```
AlanCoding: I know it's not working https://github.com/ansible/tower-cli/issues/493, but I would like internal=True on this class

Otherwise, very happy with this

love to see resources coming in with _no_ overrides.

Issue #491, opened 2018-03-14
Examples marked as "deprecated" but nothing to replace it


calston:
Guys... come on...

ryanpetrello:
@calston can you elaborate a bit?  We need some context here if we're going to help...

AlanCoding:
They are not deprecated, demonstrated by the fact that I have an open PR to maintain them right now.

What happened was that someone introduced Sphinx docs, but did not migrate the existing docs to the source tree that gets built for Read The Docs.

PR #496, opened 2018-03-16
Put docs into the actual docs site

AlanCoding:
https://github.com/ansible/tower-cli/issues/491

automatically converted md->rst, and there's probably some cleanup that it could use.

Also, I increased the max depth of the TOC, because I couldn't find anything before, on top of the fact that the titles were unhelpful.


AlanCoding:
new:

screen shot 2018-03-19 at 8 16 59 am

old:

http://tower-cli.readthedocs.io/en/latest/

matburt:
so say we all

Issue #497, opened 2018-03-19
Optionally read verify_ssl from TOWER_IGNORE_SSL


AlanCoding:
This naming was used by the tower inventory source, because of that, it is used in the tower credential type. The easiest resolution to make this work with the tower modules would be to have tower-cli read this variable if the actual environment variable, `TOWER_VERIFY_SSL`, isn't present. Ping @ryanpetrello

PR #498, opened 2018-03-19
Allow specification via alternative env var

AlanCoding:
https://github.com/ansible/tower-cli/issues/497

```
# with old var

$ TOWER_VERIFY_SSL=yes TOWER_HOST=http://foo.invalid tower-cli user list
Error: Can not verify ssl with non-https protocol. Change the verify_ssl configuration setting to continue.
$ TOWER_VERIFY_SSL=no TOWER_HOST=http://foo.invalid tower-cli user list
Error: There was a network error of some kind trying to connect to Tower.

The most common  reason for this is a settings issue; is your "host" value in `tower-cli config` correct?
Right now it is: "http://foo.invalid".

# with new var

$ TOWER_IGNORE_SSL=yes TOWER_HOST=http://foo.invalid tower-cli user list
Error: There was a network error of some kind trying to connect to Tower.

The most common  reason for this is a settings issue; is your "host" value in `tower-cli config` correct?
Right now it is: "http://foo.invalid".
$ TOWER_IGNORE_SSL=no TOWER_HOST=http://foo.invalid tower-cli user list
Error: Can not verify ssl with non-https protocol. Change the verify_ssl configuration setting to continue.
```

### DIFF
--- a/tower_cli/conf.py
+++ b/tower_cli/conf.py
@@ -40,6 +40,11 @@ CONFIG_OPTIONS = frozenset((
     'color', 'verbose', 'description_on', 'certificate',
     'use_token', 'oauth_token'
 ))
+# This mapping specifies alternative environment variables
+# which can also specify a config parameter
+ALT_ENV_OPTIONS = {
+    'verify_ssl': 'TOWER_IGNORE_SSL'
+}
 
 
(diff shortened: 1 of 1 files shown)


AlanCoding:
One second look, I don't think we really need this in _this_ repo.

Issue #507, opened 2018-03-29
Add roles to import/export 


john-westcott-iv:
Add roles to the import/export feature so that they don't need to be manually reconstructed.

john-westcott-iv:
@johnjeffers this fix should add permissions to the send/receive function. If you can please test and let me know if you have any issues.

AlanCoding:
feature is implemented

PR #508, opened 2018-03-29
Added roles to send/receive

john-westcott-iv:
Fixes for issue #507

### DIFF
--- a/tower_cli/cli/transfer/common.py
+++ b/tower_cli/cli/transfer/common.py
@@ -7,8 +7,8 @@ from tower_cli.resources.role import ACTOR_FIELDS
 ASSET_TYPE_KEY = "asset_type"
 ASSET_RELATION_KEY = "asset_relation"
 SEND_ORDER = [
-    'organization',
     'user',
+    'organization',
     'team',
     'credential_type',
     'credential',


tower_cli/cli/transfer/receive.py
```diff
@@ -104,6 +105,12 @@ def export_assets(self, all, asset_input):
                         exported_asset[common.ASSET_RELATION_KEY][relation] =\
                             common.extract_schedules(asset)['items']
 
+                # If this asset type is in the RESOURCE_FIELDS of the Role object than export its roles
+                if asset_type in RESOURCE_FIELDS:
+                    if common.ASSET_RELATION_KEY not in exported_asset:
+                        exported_asset[common.ASSET_RELATION_KEY] = {}
+                    exported_asset[common.ASSET_RELATION_KEY]['roles'] = common.extract_roles(asset)['items']
```
AlanCoding: will this copy system roles?

johnjeffers:
@john-westcott-iv Doesn't seem to be working here. This is using 1.0.4.105 as a source, loading into 1.0.5.19. Where in the export files should I expect to see the permissions data?

john-westcott-iv:
@johnjeffers The roles should be in each object exported (here is a condensed truncated credential):
```
[
    "name": "Demo Credential", 
    "asset_type": "credential", 
    "credential_type": "Machine", 
    "asset_relation": {
      "roles": [
        {
          "team": ["Team 1"], 
          "name": "Use", 
        }, 
        {
          "name": "Admin", 
          "user": ["admin", another_user"]
        }, 
      ]
    }
  }
]
```
How are you performing your build of Tower-CLI? I'm wondering if you are building the wrong branch? These changes are not on my master branch, they are on a branch called roles_io.

tower_cli/cli/transfer/receive.py
john-westcott-iv: Per our conversation, system roles won't be copied here. They will be granted at the user level.

johnjeffers:
@john-westcott-iv Yeah, I missed the branch. Working as expected now. Permissions for everything I checked were imported correctly.

tower_cli/cli/transfer/send.py
```diff
@@ -947,6 +974,95 @@ def import_schedules(self, existing_object, new_schedules, asset_type):
...
+        existing_roles = existing_role_data['items']
+        existing_name_to_id = existing_role_data['existing_name_to_object_map']
+
+        if new_roles == existing_roles:
+            self.log_ok("All roles are up to date")
+            return
+
+        existing_roles_by_name = {}
+        for role in existing_roles:
+            existing_roles_by_name[role['name']] = role
```
AlanCoding: I'm getting a little confused about name uniqueness here. AFAIK the role names (from the API) are things like "Admin", so this pivoting of a dict would lose almost all of the roles in the system, only copying memberships of 1 role of each type, which there are only a handful of.

But I'm sure there's more complex stuff going on, so basic questions about what's happening:

 - have you tested with different resource types with the same name, like test that role memberships are copied for a project named "foobar" and also copied for an inventory named "foobar" as well, and that doesn't break any of your logic?
 - are memberships in the same role type for each item copied correctly? In that example, memberships for the admin role of the project are copied distinctly from the admin role of the inventory?

PR #534, opened 2018-05-24
Updated files for RPM packaging

john-westcott-iv:
Fixed a handful of files to be able to build rpms for el6/7.
Simplified the Makefile (I think) and also added local_install for building into --user environment.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,34 +1,15 @@
 VERSION := $(shell python -c "exec(open('''tower_cli/constants.py''').read(), globals()); print VERSION")
+RELEASE := $(shell python -c "exec(open('''tower_cli/constants.py''').read(), globals()); print RELEASE")
 
-DISTS = el6 el7
-DIST_SUFFIX_el6 = 
-DIST_SUFFIX_el7 = .centos
-MOCK_CFG_el6 = epel-6-x86_64
-MOCK_CFG_el7 = epel-7-x86_64
+el6: dist/ansible-tower-cli-$(VERSION).tar.gz rpm-build/ansible-tower-cli-${VERSION}.spec
+	mock -r epel-6-x86_64 --buildsrpm --spec rpm-build/ansible-tower-cli-${VERSION}.spec --sources dist/ --resultdir rpm-build
+	mock -r epel-6-x86_64 --rebuild rpm-build/ansible-tower-cli-${VERSION}-${RELEASE}.el6.src.rpm --resultdir rpm-build
 
-.PHONY: all clean $(DISTS)
-.DEFAULT_GOAL = all
+el7: dist/ansible-tower-cli-$(VERSION).tar.gz rpm-build/ansible-tower-cli-${VERSION}.spec
+	mock -r epel-7-x86_64 --buildsrpm --spec rpm-build/ansible-tower-cli-${VERSION}.spec --sources dist/ --resultdir rpm-build
+	mock -r epel-7-x86_64 --rebuild rpm-build/ansible-tower-cli-${VERSION}-${RELEASE}.el7.src.rpm --resultdir rpm-build
 
-define RPM_DIST_RULE
-
-$(eval $(1)_SRPM=rpm-build/ansible-tower-cli-$(VERSION)-1.$(1)$(DIST_SUFFIX_$(1)).src.rpm)
-$(eval $(1)_RPM=rpm-build/ansible-tower-cli-$(VERSION)-1.$(1)$(DIST_SUFFIX_$(1)).noarch.rpm)
-SRPMS += $($(1)_SRPM)
-RPMS += $($(1)_RPM)
-
-$($(1)_SRPM): rpm-build/.exists dist/ansible-tower-cli-$(VERSION).tar.gz packaging/rpm/ansible-tower-cli.spec
-	mock -r $(MOCK_CFG_$(1)) --buildsrpm --spec packaging/rpm/ansible-tower-cli.spec --sources dist/ --resultdir rpm-build
-
-$($(1)_RPM): $($(1)_SRPM)
-	mock -r $(MOCK_CFG_$(1)) --rebuild $($(1)_SRPM) --resultdir rpm-build
-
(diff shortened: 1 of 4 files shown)


john-westcott-iv:
Woot!!! Finally code coverage increases from one of my PRs!

packaging/rpm/ansible-tower-cli.spec
```diff
@@ -47,10 +49,13 @@ rm -rf $RPM_BUILD_ROOT
...
+%doc LICENSE README.rst docs/source/HISTORY.rst
 %config(noreplace) %ghost %{_sysconfdir}/tower/tower_cli.cfg
 %{python_sitelib}/ansible*
 %{python_sitelib}/tower_cli*
+%{_bindir}/awx-cli
+%{_bindir}/tower-cli
+
 
 %changelog
 * Sun Mar 26 2017 Satoru SATOH <satoru.satoh@gmail.com> - 3.1.3-2
```
AlanCoding: ah, so this is the number 2 here.

tower_cli/constants.py
```diff
@@ -14,6 +14,8 @@
 
 
 VERSION = '3.3.0'
+# This is the release number for the RPM builds
+RELEASE = 1
```
AlanCoding: ...and the idea here is that code _importing_ tower-cli can deduce what RPM version it is.

Makefile
```diff
@@ -40,9 +21,12 @@ clean: remove_complied
 	rm -rf ansible_tower_cli.egg-info
 	rm -rf rpm-build
 
-dist/ansible-tower-cli-$(VERSION).tar.gz: HISTORY.rst LICENSE MANIFEST.in README.rst requirements.txt setup.py setup.cfg
+dist/ansible-tower-cli-$(VERSION).tar.gz: docs/source/HISTORY.rst LICENSE MANIFEST.in README.rst requirements.txt setup.py setup.cfg
```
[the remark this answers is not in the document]
john-westcott-iv: Yea, this is what started me on this little side project but I wasn't going to say anything :)

packaging/rpm/ansible-tower-cli.spec
```diff
@@ -1,14 +1,16 @@
 %{!?python_sitelib: %define python_sitelib %(%{__python} -c "from distutils.sysconfig import get_python_lib; print get_python_lib()")}
-%global srcname tower-cli
+%global srcname ansible-tower-cli
+%define version __VERSION__
+%define release __RELEASE__
 
-Name:           ansible-%{srcname}
-Version:        3.3.0
-Release:        2%{?dist}
```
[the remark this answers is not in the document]
john-westcott-iv: The version is typically the version of the software and the release is if you need to change something in the packaging of the RPM. For example, say you had a post uninstall script in the RPM that had a bug in it. The product is not changing, just the script in the RPM. So you can keep version 3.3.0 but change the release from 1 to 2.

tower_cli/constants.py
john-westcott-iv: The RPM version and release should now be controlled by setup.py. I doubt that you will need to change RELEASE often, looks like it might have only happened once in the past that I can tell.

Makefile
```diff
@@ -1,34 +1,15 @@
...
-MOCK_CFG_el7 = epel-7-x86_64
+el6: dist/ansible-tower-cli-$(VERSION).tar.gz rpm-build/ansible-tower-cli-${VERSION}.spec
+	mock -r epel-6-x86_64 --buildsrpm --spec rpm-build/ansible-tower-cli-${VERSION}.spec --sources dist/ --resultdir rpm-build
+	mock -r epel-6-x86_64 --rebuild rpm-build/ansible-tower-cli-3.3.0-1.el6.src.rpm --resultdir rpm-build
 
-.PHONY: all clean $(DISTS)
-.DEFAULT_GOAL = all
+el7: dist/ansible-tower-cli-$(VERSION).tar.gz rpm-build/ansible-tower-cli-${VERSION}.spec
+	mock -r epel-7-x86_64 --buildsrpm --spec rpm-build/ansible-tower-cli-${VERSION}.spec --sources dist/ --resultdir rpm-build
+	mock -r epel-7-x86_64 --rebuild rpm-build/ansible-tower-cli-3.3.0-1.el7.src.rpm --resultdir rpm-build
```
shanemcd: Can you update this to be something like `rpm-build/ansible-tower-cli-$(VERSION)-$(RELEASE)`?

tower_cli/constants.py
AlanCoding: > The RPM version and release should now be controlled by setup.py.

It is kind of confusing that this would be true, but it would _also_ be defined in source. It may be less prone to error to just have an overridable default value in the Makefile and leave it as that.

Makefile
john-westcott-iv: For clarification are you looking for the make target to be rpm-build/ansible-tower-cli-$(VERSION)-$(RELEASE) or the tar.gz/spec file or something else?

tower_cli/constants.py
AlanCoding: Okay, I see what you were saying now.

Reaching out to EPEL on setup seems aggressive to me. I don't want to go that far.

One thing on my mind is tags, for instance. Every initial tag, like `v3.3.2` should have `RELEASE = 1` in constants. If you ever iterated that number, it seems like you would need a tag like `v3.3.2.rev2`. Otherwise, your rpm doesn't technically correspond to a state of the source. But I don't know if that's really a problem, or something we need to care too much about.

john-westcott-iv: The other thing you could do is just never change the revision but have a change to the spec file cause a change to the version number. Then we just never have to change from revision 1 but if something is bad in the package we go from 3.3.0 to 3.3.1.

AlanCoding: True, and I would tend to prefer that. Then ideally, this would be here for the purpose of forever remaining as a static `1`.

AlanCoding:
@shanemcd are there any other changes you'd like to see here? Do you want to merge this?

PR #542, opened 2018-06-08
Fixing tower-cli execution on windows

john-westcott-iv:
os.path.sep.join was causing windows clients to use \ in a URL. This was causing calls to Tower to fail. Updated to use urljoin.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -46,12 +46,24 @@ class ClientTests(unittest.TestCase):
         with settings.runtime_values(host='33.33.33.33'):
             self.assertEqual(client.get_prefix(), 'https://33.33.33.33/api/%s/' % CUR_API_VERSION)
 
+    def test_prefix_https_protocol(self):
+        """Establish that the http protocol works
+        """
+        with settings.runtime_values(host='https://33.33.33.33'):
+            self.assertEqual(client.get_prefix(), 'https://33.33.33.33/api/%s/' % CUR_API_VERSION)
+
+    def test_prefix_http_protocol(self):
+        """Establish that the http protocol works
+        """
+        with settings.runtime_values(host='http://33.33.33.33', verify_ssl=False):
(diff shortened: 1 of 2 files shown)


AlanCoding:
This test fail looks like a meaningful find:

```python
>>> from urlparse import urljoin
>>> urljoin('bogus://33.33.33.33/', '/api/v2/')
'/api/v2/'
>>> urljoin('bogus://33.33.33.33/', '/api/')
'/api/'
```

on python 2.7.11 here.

I think we probably need to dig into the docs for how `urljoin` is expected to work.

AlanCoding:
I guess you can just change the test to use http

```
>>> urljoin('http://33.33.33.33/', 'api')
'http://33.33.33.33/api'
>>> urljoin('https://33.33.33.33/', 'api')
'https://33.33.33.33/api'
>>> urljoin('https://33.33.33.33/', '/api/')
'https://33.33.33.33/api/'
```

@ryanpetrello

john-westcott-iv:
@AlanCoding Thanks for double checking me. After looking at the source for urljoin, if the URL scheme is not "standard" the behavior we have seen is expected. 

There are two tests in api.py test:
> Establish that the prefix property returns the appropriate URL prefix given a host with no specified protocol.
and
> Establish that the prefix property returns the appropriate URL prefix and don't clobber over an explicit protocol.

Both tests were added in 2014 by @lukesneeringer . Does anyone know of a reason that TowerCLI should be allowed to use alternative schemes?

AlanCoding:
no, if you wanted to go down this road we would prefer to raise some custom exception if we got a non-standard scheme. Probably a generic exception for "your URL is wrong".

john-westcott-iv:
@AlanCoding Please see new pull. I am now throwing a ConnectionException if the scheme is not http or https. The tests now validate that http and https urls come back correct and that bogus:// throws the ConnectionException.

PR #544, opened 2018-06-09
Added --ask-tower-pass

john-westcott-iv:
This option will use click to prompt for a tower password if present. No more typing passwords on command line for your shell history to remember.

### DIFF
--- a/tower_cli/conf.py
+++ b/tower_cli/conf.py
@@ -23,6 +23,7 @@ import os
 import stat
 import warnings
 from functools import wraps
+from sys import argv
 
 import six
 from six.moves import configparser
@@ -340,6 +341,8 @@ settings = Settings()
 
 
 def _apply_runtime_setting(ctx, param, value):
+    if param.name == 'tower_password' and value == 'ASK' and '--help' not in argv:
+        value = click.prompt('Enter tower password', type=str, hide_input=True)
     settings.set_or_reset_runtime_param(param.name, value)
 
 
@@ -399,7 +402,8 @@ def with_global_options(method):
         '-p', '--tower-password',
         help='Password to use to authenticate to Ansible Tower. '
              'This will take precedence over a password provided to '
-             '`tower config`, if any.',
+             '`tower config`, if any. If value is ASK you will be '
+             'prompted for the password',
         required=False, callback=_apply_runtime_setting,
         is_eager=True
     )(method)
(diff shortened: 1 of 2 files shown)


AlanCoding:
what if, instead of a new option, we used your callback in the event that the string value of the existing password option was "ASK"? That will avoid lengthening the help text volume, and it's an approach that Tower/AWX itself has used for password prompting.

john-westcott-iv:
It would be very easy to allow prompting for all of the runtime variables. Is that something we want to do? I thought of it because I could see someone saying that want to be prompted for the username as well as the password.

tower_cli/conf.py
```diff
@@ -340,13 +340,15 @@ def config_from_environment():
...
+    if param.name == 'tower_password' and value == 'ASK':
+        value = click.prompt('Enter towwer password', type=str, hide_input=True)
     settings.set_or_reset_runtime_param(param.name, value)
 
 
 SETTINGS_PARMS = set([
     'tower_host', 'tower_oauth_token', 'tower_password', 'format',
     'tower_username', 'verbose', 'description_on', 'insecure', 'certificate',
-    'use_token'
+    'use_token', 'tower_password_ask'
```
AlanCoding: I don't know if this is needed

AlanCoding:
I would stick to password for now, it's hard to imagine a good use case for prompting for those other things.

tower_cli/conf.py
```diff
@@ -399,9 +401,10 @@ def with_global_options(method):
         '-p', '--tower-password',
         help='Password to use to authenticate to Ansible Tower. '
              'This will take precedence over a password provided to '
-             '`tower config`, if any.',
+             '`tower config`, if any. If value is ASK you will be '
+             'prompted for the password',
         required=False, callback=_apply_runtime_setting,
-        is_eager=True
+        is_eager=False
```
AlanCoding: Sorry I missed this before, but I think that `is_eager=False` won't work.

the reason is because the related lookups are also processed as field callbacks. So that means that when you do something like this:

```
tower-cli project get --organization=Default --tower-password=ASK
```

Then it looks up the related organization using the runtime settings. In the syntax I gave there, click processes the `--organization` option before the `--tower-password` option, which means that the organization will be looked up with the file settings, before the runtime settings have been applied. click supports only a single-tier of custom processing ordering, which is that eager params are processed first.

tower_cli/conf.py
```diff
@@ -340,6 +341,8 @@ def config_from_environment():
 
 
 def _apply_runtime_setting(ctx, param, value):
+    if param.name == 'tower_password' and value == 'ASK' and '--help' not in argv:
+        value = click.prompt('Enter towwer password', type=str, hide_input=True)
```
AlanCoding: towwer?

Issue #134, opened 2016-03-23
Hash password in tower-cli.cfg?


rgroten:
Currently the user password is stored plaintext in tower-cli.cfg. I'm wondering if it's be worthwhile (or even possible) to make it behave like Ansible Use module where passwords are stored in the file as hashes?

http://docs.ansible.com/ansible/faq.html#how-do-i-generate-crypted-passwords-for-the-user-module

jlaska:
@rgroten not a bad idea to obfuscate the password in case anyone is looking over your shoulder.  Good suggestion.

AlanCoding:
I would like to learn more about what the available python libraries have to offer. If this could be implemented without impact to the user experience, that would be excellent.

For some lower-hanging fruit, I wonder if we could avoid ever echoing a plaintext password.

wenottingham:
The user module password hashing is just creating the crypted string that is directly put in /etc/passwd or /etc/shadow. I'm not sure there's a direct analog of a crypted string we'd use for Tower auth unless the API supports digest auth?

rgroten:
Would be neat if Tower supported kerberos authentication.  For simple over-the-shoulder protection and keeping plaintext passwords out of files what if tower-cli prompted for the password using getpass if it's not specified in the file?

lcarva:
@rgroten It looks like Tower does support kerberos authentication: https://docs.ansible.com/ansible-tower/latest/html/administration/kerberos_auth.html. It'd be great if tower-cli supported this kind of authentication.

wenottingham:
@lcarva - that documentation is for running Tower jobs against remote systems where you'd use kerberos for authenticating connection to the Windows/Linux boxes you're connecting to.

Tower does not currently support kerberos authentication of Tower users.

lcarva:
@wenottingham got it. thanks for clarifying. sorry for the confusion.

shdobxr:
Having been a Ansible user, we are migrating to tower. I personally would like to see this feature as I would prefer not to store login's in the clear. Also was SAML2.0 ever thought? I currently generate STS tokens in AWS good for one hour. The same for tower-cli would be great as well.

MansM:
I would already be happy with just base64 version of the pass. (just like what is in your ~/.docker/config.json )

PR #545, opened 2018-06-09
Now storing password in config files as base64 encoded

john-westcott-iv:
This will base64 encode the password fields in the various config files. It will not auto-re-encode existing files but the next time config writes the password to the file it will become encoded.

### DIFF
--- a/tests/test_utils_password.py
+++ b/tests/test_utils_password.py
@@ -1,0 +1,41 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018, Ansible, Inc.
+# John Westcott <john.westcott.iv@redhat.com>
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from tower_cli.utils.password import base64_encode, base64_decode
+
+from tests.compat import unittest
+
+
+class PasswordTests(unittest.TestCase):
+    """A set of tests to establish that the encode/decode methods work
+    """
+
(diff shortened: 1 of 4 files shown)


AlanCoding:
resolves https://github.com/ansible/tower-cli/issues/134

AlanCoding:
Seems the same arguments could apply to tokens @ryanpetrello what are your thoughts?

AlanCoding:
Thinking more on this... another argument might be that this would actually be really non-standard from an Ansible perspective. Look at how we write credential files for other cloud sources in AWX:

https://github.com/ansible/awx/blob/devel/awx/plugins/inventory/cloudforms.ini.example

One of tower-cli's primary purposes is to be the upstream library for the tower_* modules, and soon we'll be adding the inventory plugin. Correspondingly, the config file that we're using now as `.tower_cli.cfg` would then likely be ported into the AWX credential creation, and no other similar examples use encryption or obfuscation. This might still work, but I would conservatively prefer to hold off until those changes have been implemented.

john-westcott-iv:
@ryanpetrello 
I don't think the intent is "Secure" I think its obfuscation and as long as we don't tout it as being secure hopefully no one is under a false sense of security. 

Many companies I know just don't want "plain text passwords" in files but they don't say they require security. At a past company someone was telling me that something to the affect of: if someone has to do something to get a password (decrypt it, unencoded it, etc) it proves intent whereas "just seeing it" and using it could be interpreted as an accident (disclaimer: I'm not a lawyer).

Also, tokens will be cool as long as people have the automation to refresh the token. i.e. if someone was using Jenkins with TowerCLI (which hopefully they are using my plugin) they would need something to rotate Jenkins token (if they set it to expiring). Also, until people have migrated to a version of Tower which supports Tokens us telling them to use a token is kind of pointless. 

@AlanCoding 
When using TowerCLI via Ansible/Tower it makes sense to use a vaulted password/Tower credential and you don't need a config file so this change would be mute. When coding it, I actually thought about using a vaulted password in the config file but then we would need to require ansible libraries on the box. I also thought about actually encrypting the password as well but didn't want to require libcrypt or another library (plus were open source so we couldn't 'hide' the decryption key). As for other files within Ansible I would say "see above". At my client they don't even keep the Tower inventory file around after building a Tower cluster because it has unencrypted passwords in the file, it is only tolerated on the filesystem for the duration of the buildout.

PR #547, opened 2018-06-13
Update application resource and documentation

willthames:
application requires an organization field, and doesn't
use the user field at all.

Add documentation of the application resource


AlanCoding:
Looks good to me, only thing that needs fixing is a minor flake8 error.

@rooftopcellist could you have a quick look, I'll be ready to merge when Travis is passing.

AlanCoding:
isn't the .rst file automatically generated?

Issue #554, opened 2018-06-18
receive, don't export hosts/groups with dynamic sources


john-westcott-iv:
When exporting an inventory which hass hosts or groups defined from a dynamic source do not export the groups or hosts.

john-westcott-iv:
@christoph-kluge sorry for the delay, we should be able to close this now.

PR #556, opened 2018-06-19
#554 changing export process to not include nodes from a sourced resource

john-westcott-iv:
If a group or host in an inventory is from a source we do not want to export the object as part of the inventory.


AlanCoding:
This will need to go in the CHANGELOG for 3.3.1

Issue #558, opened 2018-06-21
In send user roles occasionally fail to add or remove.


john-westcott-iv:
Occasionally on import we are getting message like:
    Unable to remove user XXX from Read role: The requested object could not be found.

PR #559, opened 2018-06-21
#558 Fixed issue where roles were not properly identified

john-westcott-iv:


tower_cli/cli/transfer/send.py
```diff
@@ -1144,6 +1139,19 @@ def import_roles(self, existing_asset, new_roles, asset_type):
                             actor, item_to_add, role['name'], e)
                         )
 
+    def get_role_type(self, type, name):
+        role_type = type
+        if role_type == 'Ad Hoc':
+            role_type = 'adhoc'
```
AlanCoding: Firstly, this here looks like an AWX bug. I see that you get `type` from the serialization (unless you modified it somewhere) which is

```json
            "type": "role",
```

and if you saw "Ad Hoc" there, that needs to be filed upstream to AWX. That's quite concerning.

tower_cli/cli/transfer/send.py
```diff
@@ -1144,6 +1139,19 @@ def import_roles(self, existing_asset, new_roles, asset_type):
...
+        # Some of the workflow roles come out funky so we need to straighten them out.
+        elif role_type == 'role' and name == 'Read':
+            role_type = 'read'
+        elif role_type == 'role' and name == 'Admin':
+            role_type = 'admin'
+        elif role_type == 'role' and name == 'Execute':
+            role_type = 'execute'
```
AlanCoding: This is really concerning in the tech debt area... I hate listing fields, but if I do, I want to do it once

https://github.com/ansible/tower-cli/blob/6de602d12626313eebc314efc5e4a1a4a271d65f/tower_cli/resources/role.py#L32-L36

You can pull this out of the `role.py` file, replace it with an import, then change it from a list to a dict. Once you have a dict, you can reverse the keys-values and then use that to lookup based on the `name` here.

tower_cli/cli/transfer/send.py
john-westcott-iv: The method is trying to determine the internal role name. It first assumes that you correctly passed it the internal role type (i.e. adhoc). But if you gave it "role" as the type and then the name 'Ad Hoc' it has to convert it to adhoc.

tower_cli/cli/transfer/send.py
john-westcott-iv: Since we are just converting "pretty name" to internal name (which appears to just be name.lower().replace(something) what do you think about this: 
```python
# See if we were passed the internal name directly
if role_type in ROLE_TYPES:
    return role_type
# This will catch Ad Hoc -> adhoc
if name.lower().replace(" ", "") in ROLE_TYPES:
    return name.lower().replace(" ", "")
# This would catch something like Credential Admin to credential_admin
if name.lower().replace(" ", "_") in ROLE_TYPES:
    return name.lower().replace(" ", "_")

# We couldn't find it so just raise an exception and let the caller deal
raise TowerCLIException('Role {}/{} not found'.format(role_type, name))
```

tower_cli/cli/transfer/send.py
AlanCoding: I think the API really should be providing the internal role name, but since it doesn't I'm starting to follow what you did here a little better.

tower_cli/cli/transfer/send.py
AlanCoding: this would be good for a utility method, I've been chewing on it, and the dict isn't that good of an idea.

I'm going to merge this because it fixes a bug and we'd like to have that in next release.

PR #583, opened 2018-08-05
Groups now have name fields. Helps during import.

jomeier:
Hi,

I described in this bug report that an import of inventories with groups and hosts in it will fail, because the group arrays don't have a 'name' field:

https://github.com/ansible/tower-cli/issues/582

This PR fixes that. Import of inventories with 

```
tower-cli send inventory.json
```

now works.

I would be glad if someone could check that and accept this PR.

Thanks and best regards,

Josef


jomeier:
I don't understand how one single line of code dann decrease the code coverage :-)

Issue #581, opened 2018-08-04
Export works, Import doesn't work


jomeier:
Hi,

today I tried out the export / import feature of tower-cli.

I installed the latest awx version on my machine. I started it and waited until it was accessible through the web ui. I logged in.

On the command line I exported all settings like this:
```
tower-cli receive --all > test.json
```

__test.json__ was filled with data.

Afterwards I imported this data again to the same instance of AWX like this:
```
tower-cli send test.json
```

I get this error message:
```
Asset of type user is missing identifier field username
Error: One or more errors encountered in provided assets, stopping import
```

I appended the test.json file.

Best regards,

Josef

[test.txt](https://github.com/ansible/tower-cli/files/2259554/test.txt)

jomeier:
Why are the user names not exported? 

Lot's of objects are connected to a certain user. If they are not known to tower, the import will not work at all.

john-westcott-iv:
@jomeier, This is strange behavior. In the version information you listed, is that the server you are exporting out of, importing into, or both? The user is correctly exported for me (see below) so I want to check using the same versions as you. Also, can you try to download and build tower-cli from source to see if you are still getting the same issue.

```
tower-cli receive --user admin --tower-host http://localhost:8081
[
  {
    "username": "admin", 
    "first_name": "", 
    "last_name": "", 
    "email": "admin@example.com", 
    "is_superuser": true, 
    "asset_type": "user"
  }
]
```

jomeier:
A fresh installation of AWX V1.0.7.2 shows the same behaviour. 

```
tower-cli receive --user admin
```

```
[
  {
    "asset_type": "user"
  }
]
```

lwm:
Also seeing this on a fresh `pip install ansible-tower-cli` and latest docker awx setup.

jomeier:
This is my PR which exports more user attributes. Makes the import work again.

https://github.com/ansible/tower-cli/pull/586

srgvg:
Seems like a bug in AWX API, which I myself noticed in 1.0.8 and that seems fixed in 2.0.1?

see for more details:
https://github.com/ansible/tower-cli/pull/586#issuecomment-432176155

tumluliu:
For my data migration from awx 2.0.1 to 9.1.0 with tower-cli 3.3.7, this issue still exists

TwoTwenty:
> For my data migration from awx 2.0.1 to 9.1.0 with tower-cli 3.3.7, this issue still exists

[...]

PR #586, opened 2018-08-05
Additional user information will be exported.

jomeier:
Hi,

here:
https://github.com/ansible/tower-cli/issues/581

I describe a problem, where the export of 'user' data and the reimport with the same! data doesn't work.

This PR should fix that. For some reason some required fields are not exported with 

```
tower-cli receive --user all > user.json
```

Best regards,

Josef

### DIFF
--- a/tower_cli/cli/transfer/common.py
+++ b/tower_cli/cli/transfer/common.py
@@ -48,6 +48,42 @@ API_POST_OPTIONS = {
             "required": False,
             "default": {}
         }
+    },
+    'user': {
+        "username": {
+            "required": True
+        },
+        "email": {
+            "required": True,
+            "default": ""
+        },
+        "is_superuser": {
+            "required": False,
+            "default": False
+        },
+        "is_system_auditor": {
+            "required": False,
+            "default": False
+        },
+        "first_name": {
+            "required": True,
+            "default": ""
+        },
+        "last_name": {
+            "required": True,
+            "default": ""
+        },
+        "ldap_dn": {
+            "required": False,
+            "default": ""
+        },
+        "auth": {
+            "required": False,
+            "default": []
+        },
+        "external_account": {
+            "required": False
+        }
     }
 }
 NOTIFICATION_TYPES = ['notification_templates_error', 'notification_templates_success']


lwm:
Sollid stuff @jomeier!

john-westcott-iv:
My apologies for the delay, I have been exceptionally busy over the last month. I'm not sure this is going to be the best way to fix this issue, and I think we still need to do some more troubleshooting to determine is something is going awry with your AWX installation as most users don't seem to be getting this problem.

The way the receive process is supposed to work is that TowerCLI should be asking the API about the fields for a given asset type (i.e. a user) and then determining what to output based on this. To get the required field information TowerCLI sends an HTTP OPTIONS call to the assets API end point. In the case of a user object it would be an OPTIONS call to /api/v2/users. You can do this in the UI by doing a GET on /api/v2/users and then clicking the OPTIONS button in the upper right. The result should look something like this:
```
{
    "name": "User List",
    "description": "# List Users....",
    "renders": [
        "application/json",
        "text/html"
    ],
    "parses": [
        "application/json"
    ],
    "actions": {
        "POST": {
            "username": {
                "type": "string",
                "required": true,
                "label": "Username",
                "help_text": "Required. 150 characters or fewer. Letters, digits and @/./+/-/_ only.",
                "max_length": 150
            },
            "first_name": {
                "type": "string",
                "required": false,
                "label": "First name",
                "max_length": 30
            },
           ...
```
TowerCLI then uses the POST action to decide which fields are required to build a new user (based on the required tag) and what the default value is for any optional fields.

After it gets this data, TowerCLI makes an API call to get the users themselves (via stepping through the pages of GET requests to /api/v2/users/). For each user it receives it will look at the fields of the user and match those fields to the fields in the OPTIONS call. A required field is always added to the output; an optional field whose value does not match the default value is also printed in the output.

The API_POST_OPTIONS field you are modifying was intended to handle a very specific case where performing an HTTP OPTIONS call on /api/v2/schedules does not return a POST action. By hard coding a user object in here if the definition of a user asset changed in a future version of AWX TowerCLI would have to be updated in order to handle the new fields.

So again, I think we really need to determine what exactly is going on. In the small amount of time I was able to devote to this issue I did download a couple different versions of AWX but was unable to replicate the issue. Can you post the result of an HTTP OPTIONS call for /api/v2/users? This can be done with curl using this syntax:
```curl -v -X OPTIONS http://<server>:<port>/api/v2/users/ --user <username>:<password>```

In addition, can you get the data about the admin user from the API with this curl command:
```curl -v -X GET http://<server>:<port>/api/v2/users/?username=admin --user <username>:<password>```
As stated before, technically TowerCLI will not make this call. Instead if will call /api/v2/users/ and then step through all of the pages concatenating the users into a single response to be processed by the receive command. This task would be difficult to do via curl so the above command should suffice for now.

Please be sure to run these two commands form the same machine you are running TowerCLI from. That way, if there is something between your TowerCLI and AWX installation we will capture the correct results.

As a side note, if run receive with -v it will show you messages regarding what calls it is making to AWX:
```
tower-cli receive --tower-host http://localhost:8081 --user all -v
OPTIONS http://localhost:8081/api/v2/credentials/

OPTIONS http://localhost:8081/api/v2/job_templates/

OPTIONS http://localhost:8081/api/v2/workflow_job_templates/

OPTIONS http://localhost:8081/api/v2/inventory_scripts/

OPTIONS http://localhost:8081/api/v2/projects/

OPTIONS http://localhost:8081/api/v2/credential_types/

OPTIONS http://localhost:8081/api/v2/users/

*** DETAILS: Getting records. *************************************************
GET http://localhost:8081/api/v2/users/
Params: []

OPTIONS http://localhost:8081/api/v2/teams/

OPTIONS http://localhost:8081/api/v2/organizations/

OPTIONS http://localhost:8081/api/v2/notification_templates/

OPTIONS http://localhost:8081/api/v2/inventories/

[
  {
    "username": "admin", 
    "first_name": "Admin", 
    "last_name": "User", 
    "asset_type": "user", 
    "email": "admin@awx.com"
  }, 
...
```

**Comment by TwoTwenty** [comment]
I get this with 
tower-cli receive --tower-host http://host --user all -v

WARNING: Asset type user has no API POST options no pre-checks can be performed

**Comment by srgvg** [comment]
Seems like a bug in 1.0.8 that is fixed in 2.0.1?

receiving from AWX 1.0.8:
```
➡  tower-cli receive --user all
[
  {
    "asset_type": "user"
  },
  {
    "asset_type": "user"
  }
]
```

Whilst the same (tower-cli version) on 2.0.1:
```
➡  tower-cli receive --user all
[
  {
    "asset_type": "user",
    "username": "someuser",
    "first_name": "Serge",
    "last_name": "Van Ginderachter",
    "email": "Serge.VanGinderachter@someorg.be"
  },
  {
    "asset_type": "user",
    "username": "admin",
    "first_name": "",
    "last_name": "",
    "email": "root@localhost",
    "is_superuser": true
  }
]
```

john-westcott-iv:
@TwoTwenty Thanks for the post. Are you running the export command as Admin? There may be issues where if you are running as a non-prived user you don't see the POST options.

gpduck:
I'm having the same problem with the export, however the very first time I ran the command it worked and now (minutes later) it is not working.  I've checked the User List options in my browser and it does contain the POST section within actions.  I've also tried running the curl options command and also see the POST section.  I modified my common.py to dump out the value in `return_json` and it is indeed missing the POST section.

What I've narrowed it down to is my `~/.tower_cli.cfg` file.  If I run `tower-cli config username` and `tower-cli config password`, then the export command works.  However when I run `tower-cli login MYUSER` and the config file contains a value for `oauth_token`, the export command does not work.  I ran curl using the oauth_token and it returns successfully but without the POST action.

nicolaibaralmueller:
> I'm having the same problem with the export, however the very first time I ran the command it worked and now (minutes later) it is not working. I've checked the User List options in my browser and it does contain the POST section within actions. I've also tried running the curl options command and also see the POST section. I modified my common.py to dump out the value in `return_json` and it is indeed missing the POST section.
> 
> What I've narrowed it down to is my `~/.tower_cli.cfg` file. If I run `tower-cli config username` and `tower-cli config password`, then the export command works. However when I run `tower-cli login MYUSER` and the config file contains a value for `oauth_token`, the export command does not work. I ran curl using the oauth_token and it returns successfully but without the POST action.

I have the exact same issue. Thanks for posting your findings. That worked. Tower-cli login does not work.

Issue #589, opened 2018-08-12
workflow associate_label doesn't work in v3.3.0


sumkincpp:
Used software versions are
```
Tower CLI 3.3.0
API v2
AWX 1.0.6.8
Ansible 2.5.3
```

Associating label results on undefined type for  `workflow_job_template` :
```bash
# tower-cli workflow associate_label
Usage: tower-cli workflow associate_label [OPTIONS]

  Associate a label with this workflow_job_template.

Local Options:
  --workflow_job_template WORKFLOW_JOB_TEMPLATE
                                  [required]
  --label LABEL                   [required]

Global Options:
  --use-token                     Turn on Tower's token-based authentication.
                                  No longer supported in Tower 3.3 and above.
  --certificate TEXT              Path to a custom certificate file that will
                                  be used throughout the command. Overwritten
                                  by --insecure flag if set.
  --insecure                      Turn off insecure connection warnings. Set
                                  config verify_ssl to make this permanent.
  --description-on                Show description in human-formatted output.
  -v, --verbose                   Show information about requests being made.
  -f, --format [human|json|yaml|id]
                                  Output format. The "human" format is
                                  intended for humans reading output on the
                                  CLI; the "json" and "yaml" formats provide
                                  more data, and "id" echos the object id
                                  only.
  -p, --tower-password TEXT       Password to use to authenticate to Ansible
                                  Tower. This will take precedence over a
                                  password provided to `tower config`, if any.
  -u, --tower-username TEXT       Username to use to authenticate to Ansible
                                  Tower. This will take precedence over a
                                  username provided to `tower config`, if any.
  -t, --tower-oauth-token TEXT    OAuth2 token to use to authenticate to
                                  Ansible Tower. This will take precedence
                                  over a token provided to `tower config`, if
                                  any.
  -h, --tower-host TEXT           The location of the Ansible Tower host.
                                  HTTPS is assumed as the protocol unless
                                  "http://" is explicitly provided. This will
                                  take precedence over a host provided to
                                  `tower config`, if any.

Other Options:
  --help  Show this message and exit.

# tower-cli workflow associate_label --workflow_job_template test --label test
Traceback (most recent call last):
  File "/opt/awx/venv/bin/tower-cli", line 11, in <module>
    sys.exit(cli())
  File "/opt/awx/venv/lib/python2.7/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
[...]

PR #590, opened 2018-08-15
Fix reported problems associating labels

AlanCoding:
Addresses https://github.com/ansible/tower-cli/issues/589

```
tower-cli job_template associate_label --job_template=966 --label=foo
OK. (changed: true)
tower-cli workflow associate_label --workflow=362 --label=foo
OK. (changed: false)
```


sumkincpp:
Wow, the fix is quite unexpected, but interesting!

PR #594, opened 2018-08-17
Avoid ignoring name provided while copying objects

### LINKED ISSUE #471: tower-cli ignores all overriding options when performing a copy operation
tower-cli documentation says that the **copy** operation on an object (inventory, group, host, etc) requires the [id] of the object and that any further fields are used to override the old data. However, this does not work at all. The object is copied and the new name for example contains the old name plus a timestamp instead of the new name I want.

```
tower-cli inventory copy
```
This returns the following help text.
```
Usage: tower-cli inventory copy [OPTIONS] [ID]

  Copy an inventory.

  Only the ID is used for the lookup. All provided fields are used to
  override the old data from the copied resource.

Field Options:
  -n, --name TEXT                 [REQUIRED] The name field.
  -d, --description TEXT          The description field.
  --organization ORGANIZATION     [REQUIRED] The organization field.
  --variables VARIABLES           Inventory variables, use "@" to get from
                                  file.
  --kind [|smart]                 The kind field. Cannot be modified after
                                  created.
  --host-filter TEXT              The host_filter field. Only useful when
                                  kind=smart.
  --insights-credential CREDENTIAL
                                  The insights_credential field.
...
```

So using the following command results in the new name being ignored.

```
tower-cli inventory list
== ============== ============
id      name      organization
== ============== ============
 2 test                 1
== ============== ============

tower-cli inventory copy 2 --name newinventoryname

Resource changed.
== =============== ============
id      name       organization
== =============== ============
 8 test @ 08:56:57            1
== =============== ============
```

ericzolf:
Just switching two lines to first create the default name
with timestamp and then overwrite it with the CLI options.

Closes #471

### DIFF
--- a/tower_cli/models/base.py
+++ b/tower_cli/models/base.py
@@ -667,8 +667,8 @@ class Resource(BaseResource):
 
         if new_name is None:
             # copy client-side, the old mechanism
-            newresource.update(kwargs)
             newresource['name'] = "%s @ %s" % (basename, time.strftime('%X'))
+            newresource.update(kwargs)
 
             return self.write(create_on_missing=True, **newresource)
         else:


ericzolf:
I'm not sure how the coverage can decrease so much by switching two lines, but I also got an e-mail to this regard, which seems to state something different:

> Coverage remained the same at 63.719% when pulling 63df311 on ericzolf:ericzolf-fix-copy-overwrite into 2c11587 on ansible:master

Issue #604, opened 2018-08-29
tox: Should the py26 testenv be kept ?


Spredzy:
As a potential contributor, the first thing I do when starting to work on a project is run the test suite.

When I clone `tower-cli` and run `tox` I end up with:

```
docs: commands succeeded                                                  
flake8: commands succeeded                                                
ERROR:   py26: commands failed                                            
py27: commands succeeded                                                  
py34: commands succeeded                                                  
py35: commands succeeded                                                  
py36: commands succeeded   
```

They are 208 occurrences of the same failure: `ValueError: zero length field name in format`.
(ie. `prefix = urljoin(prefix, "{}/".format(CUR_API_VERSION))`)

It happens because the format expression has been used without indices. Expression that has become valid in py27+ and py32+.

Looking at `HISTORY.rst` I can see that in the notes of `3.1.8`: `Begin process to deprecate python 2.6.`

So should the py26 tox testenv be kept?

Spredzy:
In case it shouldn't be kept I made #605 that removes it.

PR #605, opened 2018-08-29
tox: Dropping py26 testenv

Spredzy:
Dropping the tox py26 env as tests are failing with:

```
  prefix = urljoin(prefix, "{}/".format(CUR_API_VERSION))
ValueError: zero length field name in format
```

Current output of running tox:

```
  docs: commands succeeded
  flake8: commands succeeded
  ERROR:   py26: commands failed
  py27: commands succeeded
  py34: commands succeeded
  py35: commands succeeded
  py36: commands succeeded
```

Fixes: https://github.com/ansible/tower-cli/issues/604

### DIFF
--- a/hacking/README.md
+++ b/hacking/README.md
@@ -5,7 +5,7 @@ Env-setup
 ---------
 
 The 'env-setup' script modifies your environment to allow you to run
-tower-cli from a git checkout using python 2.6+.
+tower-cli from a git checkout using python 2.7+.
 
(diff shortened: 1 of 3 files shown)


AlanCoding:
We stopped running these in Travis a while ago https://github.com/ansible/tower-cli/pull/365, but we kept it as part of the tox config for running locally, but given nearly a year since that time, I agree it's high-time to remove that as well.

PR #608, opened 2018-09-12
Fix for null check on a integer type value while modifying settings

### LINKED ISSUE #600: tower-cli setting modify - can't unset / null an int type
I'm attempting to unset the int value for `LOG_AGGREGATOR_PORT` setting

```sh
# by default this setting is null
tower-cli setting get LOG_AGGREGATOR_PORT -f yaml ;
# id: LOG_AGGREGATOR_PORT
# value: null

# works fine when setting an int
tower-cli setting modify -f yaml LOG_AGGREGATOR_PORT 1 ;

# but what to do if I want to unset it?
tower-cli setting modify -f yaml LOG_AGGREGATOR_PORT ; # TypeError: int() argument must be a string or a number, not 'NoneType'
tower-cli setting modify -f yaml LOG_AGGREGATOR_PORT '' ; # ValueError: invalid literal for int() with base 10: ''
# same for 'None' 'null' '\0'
```

Apparently, the only way I have found, it through the GUI.

jainnikhil30:
Fixes #600

### DIFF
--- a/tower_cli/resources/setting.py
+++ b/tower_cli/resources/setting.py
@@ -152,7 +152,10 @@ class Resource(models.Resource):
         r = client.options(self.endpoint)
         to_type = r.json()['actions']['PUT'].get(key, {}).get('type')
         if to_type == 'integer':
-            return int(value)
+            if value != 'null':
+                return int(value)
+            else:
+                return None
         elif to_type == 'boolean':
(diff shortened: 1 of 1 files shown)


AlanCoding:
@ryanpetrello do you have any objections to this modification?

Issue #596, opened 2018-08-20
Import/Export issue with umlauts in notification template


PaulRitzkat:
### Versions
tower-cli --version: Tower CLI 3.3.0
awx version: 1.0.7

### Error
Importing notification template with umlauts in name field throws UnicodeEncodeError (i didn't test other fields with umlauts):
```
Traceback (most recent call last):
  File "/usr/bin/tower-cli", line 9, in <module>
    load_entry_point('ansible-tower-cli==3.3.0', 'console_scripts', 'tower-cli')()
  File "/usr/lib/python2.7/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python2.7/site-packages/tower_cli/cli/base.py", line 129, in invoke
    return super(TowerCLI, self).invoke(ctx)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python2.7/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/tower_cli/conf.py", line 359, in method_with_context_managed
    method(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/tower_cli/cli/misc.py", line 366, in send
    sender.send(source, prevent, secret_management)
  File "/usr/lib/python2.7/site-packages/tower_cli/cli/transfer/send.py", line 51, in send
    self.print_header_row(asset_type, asset_name)
  File "/usr/lib/python2.7/site-packages/tower_cli/cli/transfer/logging_command.py", line 27, in print_header_row
    self.my_print("{} [{}] {}".format(asset_type.replace("_", " ").upper(), asset_name, stars))
UnicodeEncodeError: 'ascii' codec can't encode character u'\xfc' in position 9: ordinal not in range(128)
```
json file from tower-cli export (names changed to protect the innocent ;) ) the name had ü in it's name:
```json
[
  {
    "notification_type": "email",
    "organization": "Default",
    "notification_configuration": {
      "username": "",
      "sender": "awx@email",
      "recipients": [
        "user@email"
[...]

PR #613, opened 2018-09-13
Make text unicode before format

AlanCoding:
Maybe resolves https://github.com/ansible/tower-cli/issues/596

I have not tested this yet.

### DIFF
--- a/tower_cli/cli/transfer/logging_command.py
+++ b/tower_cli/cli/transfer/logging_command.py
@@ -1,5 +1,6 @@
 from tower_cli.utils import secho
 import click
+import six
 
 
 class LoggingCommand:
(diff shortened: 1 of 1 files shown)


tower_cli/cli/transfer/logging_command.py
```diff
@@ -24,7 +25,9 @@ def print_header_row(self, asset_type, asset_name):
 
         # The 4 is from 2 spaces and 2 brackets
         stars = '*' * (int(self.columns) - (len(asset_type + asset_name) + 4))
-        self.my_print("{} [{}] {}".format(asset_type.replace("_", " ").upper(), asset_name, stars))
+        self.my_print(six.text_type("{} [{}] {}").format(
```
john-westcott-iv: Should we do this here or could we do this down at line 95 to cover all "my_print" calls? 
(line 96):
    message = six.text_type(message)

AlanCoding: I think we have to do it here, because it's the computation of `"{} [{}] {}".format(asset_type.replace("_", " ").upper(), asset_name, stars)` that causes the problem. There are quite a few other cases of using `.format` in this file, but I looked at all of them, and I don't see any other cases where non-ascii characters are likely to be in the arguments to it. If we pass an internal string, like type, then it's not an issue, since it's not user-defined, so it will never contain unicode.

john-westcott-iv: Cool, I'm good then.

Issue #523, opened 2018-05-04
send hardcodes black color?


wenottingham:
![screenshot from 2018-05-03 23-38-03](https://user-images.githubusercontent.com/5208768/39612184-23a41b86-4f2b-11e8-8f52-b12a127b8117.png)

It's not legible on some terminals - should this just use the default text color?

AlanCoding:
what does `ansible-playbook` do?

wenottingham:
ansible-playbook uses the default terminal color (white in the above case).

PR #614, opened 2018-09-14
#523 Changed foreground color to default, not black

john-westcott-iv:


AlanCoding:
Resolves https://github.com/ansible/tower-cli/issues/523

PR #616, opened 2018-09-18
qa: Rely on pytest rather than nosetests

Spredzy:
AWX/Tower projects rely on pytest to run their test suite.

Luckily, pytest can run nosetest perfectly fine. In order to be as
coherent as posible with the ecosystem, it might be a good idea to run
pytest also in tower-cli.

Signed-off-by: Yanis Guenane <yguenane@redhat.com>


Spredzy:
cc @AlanCoding

AlanCoding:
This doesn't work with coveralls. I wouldn't necessarily be opposed to removing that, but I need to ping someone with true administrator access to this repo because of how the integration is enabled.

PR #618, opened 2018-09-24
Checking if the field is read-only before trying to modify it

### LINKED ISSUE #551: "changed" status of setting modify command is wrong / not present
screen shot 2018-06-15 at 8 21 13 am

jainnikhil30:
Fixes #551

### DIFF
--- a/tower_cli/resources/setting.py
+++ b/tower_cli/resources/setting.py
@@ -151,6 +151,9 @@ class Resource(models.Resource):
             return json.loads(value)
         r = client.options(self.endpoint)
         to_type = r.json()['actions']['PUT'].get(key, {}).get('type')
+        if to_type is None:
+            raise exc.TowerCLIError('You are trying to modify the value of a '
+                                    'Read-Only Field, which is not allowed')
         if to_type == 'integer':
             if value != 'null':
                 return int(value)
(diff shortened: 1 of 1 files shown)


AlanCoding:
I'm going to have to write some of my own notes here. Hope you don't mind. Here is an example of a normal setting:

```json
            "LOG_AGGREGATOR_PORT": {
                "type": "integer",
                "required": false,
                "label": "Logging Aggregator Port",
                "help_text": "Port on Logging Aggregator to send logs to (if required and not provided in Logging Aggregator).",
                "category": "Logging",
                "category_slug": "logging",
                "default": null
            },
```

That setting is shown in the OPTIONS under actions -> PUT and actions -> GET

Here is an example of an unchangeable setting

```json
            "AWX_ISOLATED_KEY_GENERATION": {
                "type": "boolean",
                "label": "Generate RSA keys for isolated instances",
                "help_text": "If set, a random RSA key will be generated and distributed to isolated instances.  To disable this behavior and manage authentication for isolated instances outside of Tower, disable this setting.",
                "category": "Jobs",
                "category_slug": "jobs",
                "defined_in_file": false
            },
```

That setting is (correctly) shown under actions -> GET but not under PUT.

So, in the case that a setting is read-only, this code yields an empty dict for `r.json()['actions']['PUT'].get(key, {})`, which is why your use of `to_type` winds up being correct.

Functionally, I agree with this change, but it takes research to figure out how it works. Because of that, I would ask that you either add a descriptive comment, or change the test to `if key not in r.json()['actions']['PUT']`, for the sake of clarity. Then I'll be happy to merge this.

Thanks!

AlanCoding:
travis fail looks unrelated, will look into it

PR #659, opened 2019-01-24
Rename 'target' to 'targets' in notification template API doc

samcarpentier:
The Python API requires the `targets` field to be passed when creating an IRC notification template. 

Using the `target` field, like the doc currently proposes, causes an error to be thrown.

```json
{
  "detail": "NotificationTemplate has no field named 'target'"
}
```

Using the `targets` field solved the issue and creates the notification template successfully.


AlanCoding:
Does this need to be changed here as well?

https://github.com/ansible/tower-cli/blob/bf21df5e8a15a1796c119be8f2c21422c8f73383/tower_cli/resources/notification_template.py#L149

samcarpentier:
@AlanCoding It would appear so. There seems to be a discrepancy in the file you referenced. Both `target` and `targets` are used.

samcarpentier:
Sounds good. I will take a look at this a bit later.

Thanks for your feedback.

PR #661, opened 2019-01-28
Fix bug in processing verify_ssl default with cert

AlanCoding:


pilou-:
In addition to the bugfix, could the unit test I [provided](https://bugzilla.redhat.com/show_bug.cgi?id=1669780) be committed too ?

Is there a CVE for this issue ?

AlanCoding:
I don't have access to the link, but I would welcome a new unit test for this.

I have not created a CVE or seen one that someone else has created.

pilou-:
Here it is:
* [0001-Ensure-certificates-are-verified-by-default.patch.txt](https://github.com/ansible/tower-cli/files/2822239/0001-Ensure-certificates-are-verified-by-default.patch.txt)
* [0002-By-default-verify-certificates.patch.txt](https://github.com/ansible/tower-cli/files/2822240/0002-By-default-verify-certificates.patch.txt)

you might want to raise that your workflow doesn't seem to work.

PR #662, opened 2019-02-04
Ensure certificates are verified by default.

pilou-:
Add unit test for #661: ensure certificates are verified by default.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -108,7 +108,10 @@ class ClientTests(unittest.TestCase):
             https_adapter = client.adapters['https://']
             with mock.patch.object(FauxAdapter, 'send', wraps=https_adapter.send) as mock_send:
                 client.get('/ping/')
-                mock_send.assert_called_once()
+                mock_send.assert_called_once_with(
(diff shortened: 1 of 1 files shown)


tests/test_api.py
```diff
@@ -101,6 +102,15 @@ def test_request_post(self):
...
                              '{"payload": "this is my payload."}')
 
+    def test_connection_ssl(self):
+        with client.test_mode as t:
+            t.register_json('/ping/', {'status': 'ok'})
+            https_adapter = client.adapters['https://']
+            with mock.patch.object(FauxAdapter, 'send', wraps=https_adapter.send) as mock_send:
+                client.get('/ping/')
+                mock_send.assert_called_once()
+                self.assertTrue(mock_send.call_args[1]['verify'])
```
AlanCoding:
```python
                mock_send.assert_called_once_with(
                    mock.ANY, cert=None, proxies=mock.ANY, stream=mock.ANY,
                    timeout=mock.ANY, verify=True
                )
```

It looks like this would work with the python versions failing in Travis.

PR #673, opened 2019-03-11
Fix deep nested recursion when sending workflows

sumkincpp:
This PR fixes issue with big tower-cli asset files with lots of workflows and nodes.
At some point nodes can't be uploaded due to exceeded recursion depth.
The bug itself is due to dynamic decorator creation on every constructor call.

Error log : [node-recursion-log.txt](https://github.com/ansible/tower-cli/files/2951076/node-recursion-log.txt)

### DIFF
--- a/tower_cli/resources/node.py
+++ b/tower_cli/resources/node.py
@@ -74,9 +74,17 @@ class Resource(models.Resource):
     )
 
     def __new__(cls, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


AlanCoding:
@wenottingham do you think this is related to the introduction of workflows-in-workflows?

wenottingham:
no idea

Issue #676, opened 2019-03-15
YAMLLoadWarning when specifying extra vars


cniekerk:
When I execute `tower-cli` while specifying the `--extra-vars` parameter, I see the following warning:

```
/usr/local/lib/python2.7/dist-packages/tower_cli/utils/parser.py:94: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
  return_dict = yaml.load(var_string)
```

I am currently using:

```
ansible-tower-cli (3.3.2)
```

jainnikhil30:
Closed via #680

PR #680, opened 2019-03-21
Explicitly add Loader to yaml.load 

### LINKED ISSUE #675: pyyaml yaml.load Deprecation
https://github.com/ansible/tower-cli/blob/123bb21693bbc4d867cae1eeb1bcd59a5670d8b7/tower_cli/utils/parser.py#L94

Hi Ansible,
since PyYaml deprecreated the load function, can you fix this yaml.load() accordingly?
Otherwise, tower-cli is not working with latest python docker image with has python3.7 and the pyYaml 5.1 updated a couple weeks ago.
nb: also update pip install, please
thanks,
-Jingtao

jainnikhil30:
Fixes #675 #676

### DIFF
--- a/tests/test_resources_job.py
+++ b/tests/test_resources_job.py
@@ -157,7 +157,7 @@ class LaunchTests(unittest.TestCase):
                 result = self.res.launch(1, no_input=False)
                 self.assertDictContainsSubset(
                     {"spam": "eggs"},
-                    yaml.load(edit.mock_calls[0][1][0])
+                    yaml.load(edit.mock_calls[0][1][0], Loader=yaml.FullLoader)
                 )
             self.assertDictContainsSubset(
                 {'foo': 'bar'},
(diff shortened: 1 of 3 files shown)


tests/test_resources_job.py
```diff
@@ -178,7 +178,7 @@ def test_extra_vars_at_runtime_YAML_JT(self):
                 self.assertIn('# comment', edit.mock_calls[0][1][0])
                 self.assertDictContainsSubset(
                     {"spam": "eggs"},
-                    yaml.load(edit.mock_calls[0][1][0])
+                    yaml.load(edit.mock_calls[0][1][0], Loader=yaml.FullLoader)
```
AlanCoding: is there any reason to not make these SafeLoader?

jainnikhil30: ah my bad. I missed that. I will change that too
Thanks.

PR #693, opened 2019-04-10
change notify to boolean

### LINKED ISSUE #475: Unable to set notify true for HipChat notifications
I'm unable to use tower-cli to set a HipChat notification template to true.
I've also noticed in the help section it shows "--notify TEXT"

The error returned is:

```
Response: {"__all__":["Configuration field 'notify' incorrect type, expected bool."]}
```

jainnikhil30:
fixes #475

### DIFF
--- a/tower_cli/resources/notification_template.py
+++ b/tower_cli/resources/notification_template.py
@@ -133,8 +133,8 @@ class Resource(models.Resource):
                          'Use multiple flags to send to multiple rooms, ex '
                          '--rooms=A --rooms=B'.
                          format('hipchat'), multiple=True)
-    notify = models.Field(required=False, display=False, default=False,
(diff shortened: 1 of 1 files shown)


tower_cli/resources/notification_template.py
```diff
@@ -133,8 +133,8 @@ class Resource(models.Resource):
                          'Use multiple flags to send to multiple rooms, ex '
                          '--rooms=A --rooms=B'.
                          format('hipchat'), multiple=True)
-    notify = models.Field(required=False, display=False, default=False,
-                          help_text='[{0}]The notify channel trigger.'.
+    notify = models.Field(required=False, display=False, type=click.BOOL,
```
AlanCoding: looks like maybe we should have been using this instead of just `bool` in other cases.

Issue #672, opened 2019-03-06
tower-cli does not support username with leading zero


spjssa:
Here is an example, for valid user "008675309", which can log into the Tower web interface using the same password;
$ tower-cli credential list -u "008675309" -p ‘MyPassword'
Error: Invalid Tower authentication credentials (HTTP 401).

Other valid user, beginning with non zero succeeds ("481629847",) even when leading zero is added ("00481629847", which does not exist):

$ tower-cli credential list -u "481629847" -p 'MyPassword'
== ========================= ===============
id           name            credential_type
== ========================= ===============
71 aws-ansible-x                           5
74 ansible-vault-d                         3
42 empty-machine-credentials               1
== ========================= ===============

$ tower-cli credential list -u "00481629847" -p 'MyPassword'
== ========================= ===============
71 aws-ansible-x                           5
74 ansible-vault-d                         3
42 empty-machine-credentials               1
== ========================= ===============

It appears the username is being treated as a number somewhere along the pathway, even though the tower-cli code looks like it is a string.  Any patterns of quoting or escaping, or placing the same into .tower_cli.cfg do not make any difference.

This is a blocker for some automation we are trying to roll out, since several of the users who need to run CLI commands have the username with leading zero.  So far, we have not found any workaround for the issue.

AlanCoding:
resolved in linked PR

PR #695, opened 2019-04-15
change for using correct get method depending on param type

### LINKED ISSUE #576: .tower_cli.cfg  throws a stacktrace with a % in the password field
## Symptom
With a .tower_cli.cfg file like such:
```
[general]
username = ansibleuser
password = changme%blah
host = https://ansible.tower.com
```
tower-cli fails to parse the config file.

## Expected Behaviour
Tower uses the password successfully

## Actual Behaviour
Stacktrace
```
  File "/usr/local/bin/tower-cli", line 9, in <module>
    load_entry_point('ansible-tower-cli==3.3.0', 'console_scripts', 'tower-cli')()
  File "/usr/local/lib/python3.4/dist-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.4/dist-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.4/dist-packages/tower_cli/cli/base.py", line 129, in invoke
    return super(TowerCLI, self).invoke(ctx)
  File "/usr/local/lib/python3.4/dist-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.4/dist-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.4/dist-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.4/dist-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.4/dist-packages/tower_cli/conf.py", line 359, in method_with_context_managed
    method(*args, **kwargs)
  File "/usr/local/lib/python3.4/dist-packages/tower_cli/cli/resource.py", line 107, in func
    result = method(*args, **kwargs)
  File "/usr/local/lib/python3.4/dist-packages/tower_cli/models/base.py", line 529, in list
    response = self.read(**kwargs)
  File "/usr/local/lib/python3.4/dist-packages/tower_cli/models/base.py", line 288, in read
    r = client.get(url, params=kwargs)
  File "/usr/local/lib/python3.4/dist-packages/requests/sessions.py", line 521, in get
    return self.request('GET', url, **kwargs)
  File "/usr/local/lib/python3.4/dist-packages/tower_cli/api.py", line 217, in request
    settings.password,
  File "/usr/local/lib/python3.4/dist-packages/tower_cli/conf.py", line 217, in __getattr__
    value = parser.get('general', key)
  File "/usr/lib/python3.4/configparser.py", line 773, in get
    d)
  File "/usr/lib/python3.4/configparser.py", line 374, in before_get
    self._interpolate_some(parser, option, L, value, section, defaults, 1)
  File "/usr/lib/python3.4/configparser.py", line 423, in _interpolate_some
    "found: %r" % (rest,))
configparser.InterpolationSyntaxError: '%' must be followed by '%' or '(', found: "%blah'"

```

jainnikhil30:
fixes #576 #672

### DIFF
--- a/tower_cli/conf.py
+++ b/tower_cli/conf.py
@@ -42,7 +42,7 @@ CONFIG_PARAM_TYPE = {
     'password': click.STRING,
     'verify_ssl': click.BOOL,
     'format': click.Choice,
-    'color': click.STRING,
+    'color': click.BOOL,
     'verbose': click.BOOL,
     'description_on': click.BOOL,
     'certificate': click.STRING,
@@ -233,23 +233,19 @@ class Settings(object):
             try:
                 if CONFIG_PARAM_TYPE[key] == (click.STRING or click.Choice):
                     value = getattr(parser, configparser.ConfigParser.get.__name__)('general', key)
-                    break
                 elif CONFIG_PARAM_TYPE[key] == click.BOOL:
                     value = getattr(parser, configparser.ConfigParser.getboolean.__name__)('general', key)
-                    break
                 elif CONFIG_PARAM_TYPE[key] == click.FLOAT:
                     value = getattr(parser, configparser.ConfigParser.getfloat.__name__)('general', key)
-                    break
                 elif CONFIG_PARAM_TYPE[key] == click.INT:
                     value = getattr(parser, configparser.ConfigParser.getint.__name__)('general', key)
-                    break
             except ValueError:
                 click.secho('Value for %s is not in expected type' % key)
 
-        # Write the value to the cache, so we don't have to do this lookup
-        # logic on subsequent requests.
-        self._cache[key] = value
-        return self._cache[key]
+            # Write the value to the cache, so we don't have to do this lookup
+            # logic on subsequent requests.
+            self._cache[key] = value
+            return self._cache[key]
 
         # If we got here, that means that the attribute wasn't found, and
         # also that there is no default; raise an exception.
--- a/tower_cli/conf.py
+++ b/tower_cli/conf.py
@@ -231,7 +231,7 @@ class Settings(object):
 
             # We have a value; try to get its type and return it accordingly
             try:
-                if CONFIG_PARAM_TYPE[key] == (click.STRING or click.Choice):
+                if CONFIG_PARAM_TYPE[key] == click.STRING or CONFIG_PARAM_TYPE[key] == click.Choice:
                     value = parser.get('general', key)
                 elif CONFIG_PARAM_TYPE[key] == click.BOOL:
                     value = parser.getboolean('general', key)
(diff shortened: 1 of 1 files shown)


tower_cli/conf.py
```diff
@@ -219,21 +229,27 @@ def __getattr__(self, key):
...
+                click.secho('Value for %s is not in expected type' % key)
 
-            # Write the value to the cache, so we don't have to do this lookup
-            # logic on subsequent requests.
-            self._cache[key] = value
-            return self._cache[key]
+        # Write the value to the cache, so we don't have to do this lookup
+        # logic on subsequent requests.
+        self._cache[key] = value
+        return self._cache[key]
```
AlanCoding: Some of the tests are failing because of this change in indentation, because it looks like we were relying on the `continue` to avoid these lines in the event that a variable was not defined in any of the available parsers. Easiest solution for you might be to just re-indent them.

jainnikhil30: @AlanCoding Yeah and I think I will also have to remove those breaks. Otherwise it wont ever write a value to cache if I re-indent it.

AlanCoding: oh, I missed that.

Previously the `break` statement was breaking out of the `for tm in type_method` loop, and now it's breaking out of the `for parser in self._parsers` loop. I agree you just can't keep using the break statement in the same way. Like you say, removing the break might work fine, because you're using `elif` for the other conditions.

jainnikhil30: @AlanCoding Yeah. My bad. I should have run those integration tests locally. I have made the change and ran them locally, all green now. Thanks for your review.

tower_cli/conf.py
```diff
@@ -219,16 +229,18 @@ def __getattr__(self, key):
...
-            type_method = ('getint', 'getfloat', 'getboolean')
-            for tm in type_method:
-                try:
-                    value = getattr(parser, tm)('general', key)
-                    break
-                except ValueError:
-                    pass
+            # We have a value; try to get its type and return it accordingly
+            try:
+                if CONFIG_PARAM_TYPE[key] == (click.STRING or click.Choice):
```
AlanCoding: I expect that this won't work for the case of `click.Choice`

```python
>>> 2 == (1 or 2)
False
```

Reason is because

```python
>>> (1 or 2)
1
```

So you have to put the `or` between equality comparisons, like `2 == 1 or 2 == 2`, which would return True.

Alternatively, you can do `2 in (1, 2)`

jainnikhil30: @AlanCoding Yup. Bad use of logical operators from me :(. Corrected that.

Issue #716, opened 2019-07-01
awx-cli send is failing when job templates have machine credentials


MarkRoddick:
Tower CLI 3.3.4
AWX 5.0.0.0

### awx-cli send
The awx-cli send is returning an error when trying to send a job template with machine credentials:

```
bash-4.2# awx-cli send test.json

JOB TEMPLATE [test] ************************************************************************************************************************************************************************
Option credential is not a valid option for job_template test

PLAY RECAP *********************************************************************************************************************************************************************************
    ok=0    changed=0    warnings=0    failed=1

bash-4.2# cat test.json
[
  {
    "name": "test",
    "asset_type": "job_template",
    "project": "ansible_infra",
    "inventory": "localhost",
    "credential": "localhost",
    "asset_relation": {
       "extra_credentials": [
        "ansible"
      ],
      "notification_templates_error": [],
      "notification_templates_success": [],
      "roles": [
        {
          "team": [],
          "name": "Admin",
          "user": [
            "admin"
          ]
        },
        {
          "team": [],
          "name": "Execute",
          "user": []
        },
        {
          "team": [],
          "name": "Read",
          "user": []
        }
      ],
      "labels": [],
      "schedules": [],
      "survey_spec": {}
    },
    "playbook": "configure-reverse-proxy.yml"
  }
]

```
### awx-cli receive
The awx-cli receive command is not pulling back any machine credentials:

```
bash-4.2# awx-cli receive --job_template test
[
  {
    "name": "test",
    "asset_type": "job_template",
    "project": "ansible_infra",
    "inventory": "localhost",
    "asset_relation": {
      "extra_credentials": [
        "ansible"
      ],
      "notification_templates_error": [],
      "notification_templates_success": [],
      "roles": [
        {
          "team": [],
          "name": "Admin",
          "user": [
            "admin"
          ]
        },
        {
          "team": [],
          "name": "Execute",
          "user": []
        },
        {
          "team": [],
          "name": "Read",
          "user": []
        }
      ],
      "labels": [],
      "schedules": [],
      "survey_spec": {}
    },
    "playbook": "configure-reverse-proxy.yml"
  }
]
```

![image](https://user-images.githubusercontent.com/19282460/60441591-34747280-9c0f-11e9-9093-a6497517b3a9.png)

mpunix:
I am seeing the same error while migrating the job_template from AWX 4.0.0 to 5.0.0. 

Option credential is not a valid option for job_template

[...]

PR #610, opened 2018-09-12
Support for multi-credential / multi-vault

AlanCoding:
Addresses https://github.com/ansible/tower-cli/issues/404 and https://github.com/ansible/tower-cli/issues/525

PR #717, opened 2019-07-03
Fix to tower-cli to send and receive ALL credentials

MarkRoddick:
As the `/credential` api has been deprecated (https://github.com/ansible/awx/pull/3413) the `tower-cli "send | receive"` is not processing any machine credentials https://github.com/ansible/tower-cli/issues/716.

This PR deprecates the `extra_credentials` from the `send `and `receive `commands and utilizes the `/credentials` api to send and receive all credentials.

```
bash-4.2# tower-cli receive --job_template test
[
  {
    "name": "test",
    "allow_simultaneous": true,
    "asset_relation": {
      "notification_templates_error": [
        "Webhook"
      ],
      "notification_templates_success": [],
      "roles": [
        {
          "team": [],
          "name": "Admin",
          "user": [
            "admin"
          ]
        },
        {
          "team": [],
          "name": "Execute",
          "user": []
        },
        {
          "team": [],
          "name": "Read",
          "user": []
        }
      ],
      "labels": [],
      "schedules": [],
      "credentials": [
        "Machine - Linux",
        "Storage Account Keys"
      ],
      "survey_spec": {}
    },
    "project": "test_project",
    "inventory": "test_inventory",
    "asset_type": "job_template",
    "playbook": "playbook.yml"
  }
]
```

### DIFF
--- a/tower_cli/cli/transfer/send.py
+++ b/tower_cli/cli/transfer/send.py
@@ -284,8 +284,8 @@ class Sender(LoggingCommand):
                             self.import_inventory_groups(existing_object, relations[a_relation])
                         elif a_relation in common.NOTIFICATION_TYPES:
                             self.import_notification_relations(existing_object, relations[a_relation], a_relation)
-                        elif a_relation == 'extra_credentials':
-                            self.import_extra_credentials(existing_object, relations[a_relation])
+                        elif a_relation == 'credentials':
+                            self.import_credentials(existing_object, relations[a_relation])
                         elif a_relation == 'schedules':
                             self.import_schedules(existing_object, relations[a_relation], asset_type)
                         elif a_relation == 'roles':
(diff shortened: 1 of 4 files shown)


AlanCoding:
Thank you for your contribution.

I haven't fully digested this diff, but I can say this - this diff will require us to declare a checkpoint version for people doing upgrades. There was a point where the endpoint "credentials" did not exist, and if the user _receives_ from that version of AWX or Tower, then it will not work correctly.

That is sensible to me (in addition to the fact it doesn't work right now). I don't have the version numbers in-hand right now, but we could find these as we prepare for the next release.

tower_cli/cli/transfer/send.py
```diff
@@ -935,23 +935,23 @@ def import_extra_credentials(self, existing_object, new_creds):
...
         for cred in list(set(new_creds).difference(existing_creds)):
             try:
                 new_credential = tower_cli.get_resource('credential').get(**{'name': cred})
             except TowerCLIError as e:
-                self.log_error("Unable to resolve extra credential {} : {}".format(cred, e))
+                self.log_error("Unable to resolve credential {} : {}".format(cred, e))
                 continue
 
             try:
                 tower_cli.get_resource('job_template').associate_credential(existing_object['id'], new_credential['id'])
```
AlanCoding: It looks like `associate_credential` still uses the extra_credentials endpoint.

https://github.com/ansible/tower-cli/blob/ed8546b88fab6b07566fa1c2dc3385c82aa85e34/tower_cli/resources/job_template.py#L145

I would expect that https://github.com/ansible/tower-cli/pull/610 would still be necessary for this to work

AlanCoding:
I think I'm actually okay with this, assuming that #610 is also merged, and maybe #734 too. If #734 is not brought into play, then some of the entries in the dependencies list may need to be removed, or else they could not be sent to upgraded versions. Ping @ryanpetrello

AlanCoding:
Thanks for the contribution, I did integrate some stuff from your work here into https://github.com/ansible/tower-cli/pull/741, but I'm also working on integrating that into the AWX CI system and doing some mild reverse compatibility. So I'm closing this in favor of that.

Issue #706, opened 2019-05-23
import of job_template schedules fails when survey present


nicolaibaralmueller:
So, import of schedules fails when the schedule has survey data instead of extra vars.

Schedule with prompt for inventory also missing which inventory.

Export/Import really needs improvements!

AlanCoding:
merged for 3.3.5 release

PR #718, opened 2019-07-03
#706 Fixing issue when importing a job template with a schedule.

john-westcott-iv:


AlanCoding:
this is going to be fixing #706, and I'm going to be marking that for the next release

Issue #719, opened 2019-07-08
job template creation with associated credentials since AWX 5.0


tumbl3w33d:
I was about to create a bug in AWX until I found [this reply](https://github.com/ansible/awx/issues/4148#issuecomment-504763528) which shows it's expected behavior, but right now the ansible module `tower_job_template`, which in turn uses `tower-cli` as far as I can see, cannot create a job template with credential association anymore (it worked up until including AWX 4.0).

As stated in the linked comment, two requests are necessary, so either `tower-cli` should wrap that or the ansible module should handle this. What is your opinion on this or did I miss some progress in one of the related pieces that would tackle this issue?

There is no error on ansible module execution and the credential name is sent with the request as it used to. The job template gets created, however, it is not associated with any credentials.

tumbl3w33d:
cc @wwitzel3 as maintainer of the ansible module

AlanCoding:
I don't think it would be prudent to get rid of this field, because we still have to support old versions.

https://github.com/ansible/tower-cli/blob/7b7d3a8cc991e1c4fb4cb5bc4e4c1f251f52daa0/tower_cli/resources/job_template.py#L44

But maybe some post-creation or post-modification checking that all fields have been set might help people out. But ultimately it's the modules that need to change in some way or another.

Klaas-:
@AlanCoding I think he means tower-cli no longer creates job_templates with credentials:

`tower-cli job_template create --name tower-cli-test --project my_project --playbook my_playbook.yml --credential my_credential --inventory my_inventory`

will come back as a success but will not have a credential associated.

AlanCoding:
This situation is the same for both --credential and --vault-credential.

Klaas-:
didn't recognize that, sorry. I was just puzzled why you talk about vault_credentials :)

Klaas-:
For anyone else stumbling over this issue - as a workaround I now use rest_get and uri modules (https://github.com/redhat-cop/infra-ansible/blob/master/roles/ansible/tower/library/rest_get.py)

My workflow looks like this:
rest_get credentials, look up credential id, rest_get credentials of template, check if ssh credential already associated, remove if wrong credential, add credential unless right one already defined

adpavlov:
> For anyone else stumbling over this issue - as a workaround I now use rest_get and uri modules (https://github.com/redhat-cop/infra-ansible/blob/master/roles/ansible/tower/library/rest_get.py)
> 
> My workflow looks like this:
[...]

PR #720, opened 2019-07-10
Introduce field status to error if server is post-removal

AlanCoding:
A proposed solution to make https://github.com/ansible/tower-cli/issues/719 just a little less bad

```
tower-cli job_template create --name="Hello World Debug" --description="debug statement" --inventory=localhost --credential=user2 --project="Ansible Examples" --playbook=debug.yml -v
*** DETAILS: The inventory field is given as a name; looking it up. ***********
GET https://localhost:8043/api/v2/inventories/
Params: [('name', 'localhost')]

*** DETAILS: The credential field is given as a name; looking it up. **********
GET https://localhost:8043/api/v2/credentials/
Params: [('name', 'user2')]

*** DETAILS: The project field is given as a name; looking it up. *************
GET https://localhost:8043/api/v2/projects/
Params: [('name', 'Ansible Examples')]

GET https://localhost:8043/api/v2/job_templates/
Params: [('name', u'Hello World Debug')]

POST https://localhost:8043/api/v2/job_templates/
Data: {'job_type': u'run', 'playbook': u'debug.yml', 'description': u'debug statement', 'inventory': 2, 'credential': 2, 'name': u'Hello World Debug', 'project': 8}

Due to field error, attempting to delete created item.
DELETE /job_templates/15/
DELETE https://localhost:8043/api/v2/job_templates/15/

Error: The provided fields credential vault_credential were removed in an AWX or Tower release and failed to take effect here.
```

```
tower-cli job_template modify 17 --credential=1
Error: The provided fields credential vault_credential were removed in an AWX or Tower release and failed to take effect here.
```

```
tower-cli job_template create
Usage: tower-cli job_template create [OPTIONS]

  Create a job template.

  Fields in the resource's --identity tuple are used for a lookup; if a
  match is found, then no-op (unless --force-on-exists is set) but do not
  fail (unless --fail-on-found is set).

Field Options:
  -n, --name TEXT                 [REQUIRED] The name field.
  -d, --description TEXT          The description field.
  --job-type [run|check]          The job_type field.
  -i, --inventory INVENTORY       The inventory field.
  --project PROJECT               [REQUIRED] The project field.
  --playbook TEXT                 [REQUIRED] The playbook field.
  --credential CREDENTIAL         [DEPRECATED] The credential field.
  --vault-credential CREDENTIAL   [DEPRECATED] The vault_credential field.
  --forks INTEGER                 The forks field.
  --limit TEXT                    The limit field.
  --verbosity [default|verbose|more_verbose|debug|connection|winrm]
                                  The verbosity field.
  -e, --extra-vars VARIABLES      Extra variables used by Ansible in YAML or
                                  key=value format. Use @ to get YAML from a
                                  file.
  --job-tags TEXT                 The job_tags field.
  --custom-virtualenv TEXT        The custom_virtualenv field.
  --force-handlers BOOLEAN        The force_handlers field.
  --skip-tags TEXT                The skip_tags field.
  --start-at-task TEXT            The start_at_task field.
  --timeout INTEGER               The amount of time (in seconds) to run
                                  before the task is canceled.
  --use-fact-cache BOOLEAN        If enabled, Tower will act as an Ansible
                                  Fact Cache Plugin; persisting facts at the
                                  end of a playbook run to the database and
                                  caching facts for use by Ansible.
  --host-config-key TEXT          Allow Provisioning Callbacks using this host
                                  config key
  --ask-diff-mode-on-launch BOOLEAN
                                  Ask diff mode on launch.
  --ask-variables-on-launch BOOLEAN
                                  Prompt user for extra_vars on launch.
  --ask-limit-on-launch BOOLEAN   Prompt user for host limits on launch.
  --ask-tags-on-launch BOOLEAN    Prompt user for job tags on launch.
  --ask-skip-tags-on-launch BOOLEAN
                                  Prompt user for tags to skip on launch.
  --ask-job-type-on-launch BOOLEAN
                                  Prompt user for job type on launch.
  --ask-verbosity-on-launch BOOLEAN
                                  Prompt user for verbosity on launch.
  --ask-inventory-on-launch BOOLEAN
                                  Prompt user for inventory on launch.
  --ask-credential-on-launch BOOLEAN
                                  Prompt user for machine credential on
                                  launch.
  --survey-enabled BOOLEAN        Prompt user for job type on launch.
  --become-enabled BOOLEAN        The become_enabled field.
  --diff-mode BOOLEAN             If enabled, textual changes made to any
                                  templated files on the host are shown in the
                                  standard output.
  --allow-simultaneous BOOLEAN    The allow_simultaneous field.
  --survey-spec VARIABLES         On write commands, perform extra POST to the
                                  survey_spec endpoint.

Local Options:
  --fail-on-found    If used, return an error if a matching record already
                     exists.  [default: False]
  --force-on-exists  If used, if a match is found on unique fields, other
                     fields will be updated to the provided values. If False,
                     a match causes the request to be a no-op.  [default:
                     False]

Global Options:
  --use-token                     Turn on Tower's token-based authentication.
                                  No longer supported in Tower 3.3 and above.
  --certificate TEXT              Path to a custom certificate file that will
                                  be used throughout the command. Overwritten
                                  by --insecure flag if set.
  --insecure                      Turn off insecure connection warnings. Set
                                  config verify_ssl to make this permanent.
  --description-on                Show description in human-formatted output.
  -v, --verbose                   Show information about requests being made.
  -f, --format [human|json|yaml|id]
                                  Output format. The "human" format is
                                  intended for humans reading output on the
                                  CLI; the "json" and "yaml" formats provide
                                  more data, and "id" echos the object id
                                  only.
  -p, --tower-password TEXT       Password to use to authenticate to Ansible
                                  Tower. This will take precedence over a
                                  password provided to `tower config`, if any.
                                  If value is ASK you will be prompted for the
                                  password
  -u, --tower-username TEXT       Username to use to authenticate to Ansible
                                  Tower. This will take precedence over a
                                  username provided to `tower config`, if any.
  -t, --tower-oauth-token TEXT    OAuth2 token to use to authenticate to
                                  Ansible Tower. This will take precedence
                                  over a token provided to `tower config`, if
                                  any.
  -h, --tower-host TEXT           The location of the Ansible Tower host.
                                  HTTPS is assumed as the protocol unless
                                  "http://" is explicitly provided. This will
                                  take precedence over a host provided to
                                  `tower config`, if any.

Other Options:
  --help  Show this message and exit.
```

### DIFF
--- a/tower_cli/models/base.py
+++ b/tower_cli/models/base.py
@@ -423,6 +423,20 @@ class BaseResource(six.with_metaclass(ResourceMeta)):
         # Actually perform the write.
         r = getattr(client, method.lower())(url, data=kwargs)
 
+        # Check for deprecated fields that failed to take effect
+        failed_fields = []
+        for field in self.fields:
+            if field.deprecated and kwargs.get(field.name) and field.name not in r:
+                # Uh oh, user using field that have been removed
+                failed_fields.append(field)
+        if failed_fields:
+            if method == 'POST':
+                debug.log('Due to field error, will attempt to delete created item.')
+                self.delete(r.json()['id'])
+            raise exc.TowerCLIError(
+                'The provided fields %s were removed in an AWX or Tower release '
+                'and failed to take effect here.' % ' '.join([f.name for f in failed_fields]))
+
         # At this point, we know the write succeeded, and we know that data was changed in the process.
         answer = OrderedDict((('changed', True), ('id', r.json()['id'])))
         answer.update(r.json())
--- a/tests/test_resources_job_template.py
+++ b/tests/test_resources_job_template.py
@@ -41,7 +41,7 @@ class TemplateTests(unittest.TestCase):
             t.register_json(endpoint, {'changed': True, 'id': 42},
                             method='POST')
             self.res.create(name='bar', job_type='run', inventory=1,
-                            project=1, playbook='foobar.yml', credential=1)
+                            project=1, playbook='foobar.yml')
             self.assertEqual(t.requests[0].method, 'GET')
             self.assertEqual(t.requests[1].method, 'POST')
             self.assertEqual(len(t.requests), 2)
@@ -55,7 +55,7 @@ class TemplateTests(unittest.TestCase):
             t.register_json(endpoint, {'changed': True, 'id': 42},
                             method='POST')
             self.res.create(name='bar', inventory=1, project=1,
-                            playbook='foobar.yml', credential=1)
+                            playbook='foobar.yml')
             req_body = json.loads(t.requests[1].body)
             self.assertIn('job_type', req_body)
             self.assertEqual(req_body['job_type'], 'run')
@@ -74,13 +74,13 @@ class TemplateTests(unittest.TestCase):
                                 'playbook': 'foobar.yml', 'credential': 1},
                             method='POST')
             self.res.create(name='bar', job_type='run', inventory=1,
-                            project=1, playbook='foobar.yml', credential=1)
+                            project=1, playbook='foobar.yml')
 
             f = ResSubcommand(self.res)._echo_method(self.res.create)
             with mock.patch.object(click, 'secho'):
                 with settings.runtime_values(format='human'):
                     f(name='bar', job_type='run', inventory=1,
-                      project=1, playbook='foobar.yml', credential=1)
+                      project=1, playbook='foobar.yml')
 
     def test_create_w_extra_vars(self):
         """Establish that a job template can be created
@@ -94,7 +94,7 @@ class TemplateTests(unittest.TestCase):
             t.register_json(endpoint, {'changed': True, 'id': 42},
                             method='POST')
             self.res.create(name='bar', job_type='run', inventory=1,
-                            project=1, playbook='foobar.yml', credential=1,
+                            project=1, playbook='foobar.yml',
                             extra_vars=['foo: bar'])
             self.assertEqual(t.requests[0].method, 'GET')
             self.assertEqual(t.requests[1].method, 'POST')
--- a/tower_cli/cli/resource.py
+++ b/tower_cli/cli/resource.py
@@ -353,6 +353,8 @@ class ResSubcommand(click.MultiCommand):
                     option_help += ' Use @ to get JSON or YAML from a file.'
                 if field.required:
                     option_help = '[REQUIRED] ' + option_help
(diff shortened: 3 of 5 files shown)


tower_cli/models/base.py
```diff
@@ -423,6 +423,20 @@ def write(self, pk=None, create_on_missing=False, fail_on_found=False, force_on_
...
+        # Check for deprecated fields that failed to take effect
+        failed_fields = []
+        for field in self.fields:
+            if field.deprecated and kwargs.get(field.name) and field.name not in r:
+                # Uh oh, user using field that have been removed
+                failed_fields.append(field)
+        if failed_fields:
+            if method == 'POST':
+                debug.log('Due to field error, will attempt to delete created item.')
+                self.delete(r.json()['id'])
```
ryanpetrello: I'm not sure I understand the intent here.  If they specified a deprecated field, delete the thing they just made?  This seems like fairly surprising behavior.

AlanCoding: > If they specified a deprecated field, delete the thing they just made?

There are 2 conditions that have to be met:

 - they specified a deprecated field
 - that deprecated field they specified failed to take effect

There is a very tense decision for the client to make - warning or error. Continuing without errors is going to cause havoc. But if the user sees an error, they would hope that no changes were committed. This is the closest we can get to some form of atomicity. Erring seems right to me. A warning would set the expectation that the object still exists, but would allow execution to continue. All options are messy, this is the best I could come up with.

ryanpetrello: Yea, I don't think I like this behavior.  It's quite different from what, say, the `/api/` browser client does, and I think it's fairly surprising (maybe even more surprising than the field you specified not working).

ryanpetrello: Maybe for fields like this that were deprecated which we removed, we should actually update the Tower API to detect when they're provided and raise an `HTTP 400` that explicitly describes that field no longer works.

AlanCoding: > It's quite different from what, say, the /api/ browser client does

It does what it does for REST principles, not usability. Therein lies the assumption that the client deals with problems like these (it has OPTIONS available, etc.), and this is the client we're talking about.

I would go even further in the direction of this PR, to wonder if maybe this ex post facto check should be universal in the client, for all fields it attempts to write. It's was likely to matter since tower-cli directly encodes the schema in code, but that changed with these fields.

PR #721, opened 2019-07-17
Fix 'Found more than one unified job' bug

MarBra:
Hi :)

the send command can't find the right job, if you setup a job and a workflow with the same name.

Steps to reproduce:
- Create job a
- Create workflow a
- Create workflow b
  - Include Job a

`tower-cli receive --workflow b > b.json`
`tower-cli send b.json`

Cheers
Marcel


john-westcott-iv:
If we set it starts_with and there is another job_type_x added later that will pick it up instead of not finding those if we do an exact match. So I am good to merge this in. Thanks @MarBra for the fix!

PR #733, opened 2019-08-23
Fixed using worflow-template name instead of primary key

### LINKED ISSUE #701: workflow_job.launch: workflow_job_template cannot use name
Hi,

`launch` method of `workflow_job` does not work with the `name` of the workflow job template, but it's OK with its `Primary key`.

Regards
Alexis

dnagl:
Fixes #701


AlanCoding:
Per the comment left in the issue, I think this is a matter of documenting the syntax

PR #738, opened 2019-09-12
Redirect to new AWX CLI

AlanCoding:
Screen Shot 2019-09-11 at 10 22 13 PM

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -4,22 +4,29 @@
 Welcome to tower-cli
 ====================
 
-**tower-cli** is a command line tool for Ansible Tower. It allows Tower
-commands to be easily run from the Unix command line. It can also be
-used as a client library for other python apps, or as a reference for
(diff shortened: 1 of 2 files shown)


ktdreyer:
Will the `tower_*` modules migrate to this newer library?

I'm not even sure where those modules are hosted in Git anymore with the module reorganization in v2.10.0.

AlanCoding:
> I'm not even sure where those modules are hosted in Git anymore with the module reorganization in v2.10.0.

It's in a new-style collection at https://galaxy.ansible.com/awx/awx

They no longer require tower-cli, but do the same thing.

ktdreyer:
Thank you for that link @AlanCoding !

I found a small bug with "ping" in `module_utils/ansible_tower.py` that I want to fix, so I will try reproducing this with the master branch there.

PR #741, opened 2019-10-24
Get send/receive to work with AWX 8.0.0 and JT credentials

AlanCoding:
Closes https://github.com/ansible/tower-cli/issues/716

Yanked the JT code from https://github.com/ansible/tower-cli/pull/610

This is so that the `tower_send` and `tower_receive` modules become self-consistent (creates the data it receives) for the AWX collection version 8.0.0+


AlanCoding:
Now replaces https://github.com/ansible/tower-cli/pull/734/commits as well.

PR #750, opened 2020-03-12
Improve error handling for users using the schema command

AlanCoding:
It looks like the server stopped verifying that `unified_job_template` has a non-null value because of the additional of approval nodes.

This has a consequence for function of the schema command, since it was never documented how to use workflows-in-workflows. It would silently finish with null nodes, which is a sideways state.

### DIFF
--- a/docs/source/cli_ref/usage/WORKFLOWS.rst
+++ b/docs/source/cli_ref/usage/WORKFLOWS.rst
@@ -127,6 +127,16 @@ network with a tower-cli command after the constituent resources like
 the job templates and projects were created by preceding tower-cli
 commands.
 
+Workflows Inside Workflows
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+It is possible to have a workflow embedded inside of the schema of another
+workflow. To do this, use the key "workflow".
+
+.. code:: yaml
+
+    - workflow: Another workflow
+
 Differences with Machine Formatted Schemas
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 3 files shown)


tower_cli/resources/workflow.py
```diff
@@ -45,16 +45,19 @@ def __init__(self, data, wfjt, include_id=False):
             else:
                 node_attrs[fd] = data[fd]
         node_attrs['workflow_job_template'] = wfjt
+        ujt_ambiguity_msg = (
+            'You should provide exactly one of the attributes'
+            ' job_template, project, workflow, inventory_source, or unified_job_template.'
+        )
```
beeankha: This is way better wording!

tower_cli/resources/workflow.py
```diff
@@ -45,16 +45,19 @@ def __init__(self, data, wfjt, include_id=False):
             else:
                 node_attrs[fd] = data[fd]
         node_attrs['workflow_job_template'] = wfjt
+        ujt_ambiguity_msg = (
+            'You should provide exactly one of the attributes'
+            ' job_template, project, workflow, inventory_source, or unified_job_template.'
```
kedark3: :+1:  for this clear message. Although I am not familiar with `unified_job_template`

docs/source/cli_ref/usage/WORKFLOWS.rst
```diff
@@ -127,6 +127,16 @@ network with a tower-cli command after the constituent resources like
 the job templates and projects were created by preceding tower-cli
 commands.
 
+Workflows Inside Workflows
```
kedark3: This is good. How about adding `workflow` above as well in the example https://github.com/ansible/tower-cli/pull/750/files#diff-e888d13496218a3618bbee8be0d2a7d1R114-R124 ? I see example already shows jt, project, inv_source, so we can add `workflow` there and make it a complete example. Just a suggestion.

AlanCoding: I hope you don't mind that I don't do that out of sheer laziness. I _know_ that I've tested the content here is functional. The implication is that you can combine them all together, but we just don't want poke this library any more than we have to.

PR #751, opened 2020-03-12
pin the click library to fix tests

AlanCoding:


ktdreyer:
Unfortunately Fedora 30 and above have Click v7. https://koji.fedoraproject.org/koji/packageinfo?packageID=18578

